Issue #3, opened 2018-06-22
Initialize reason and sqlState in PostgresqlServerErrorException


mp911de:
`PostgresqlServerErrorException` should initialize reason and sqlState of its `R2dbcException` superclass.

nebhale:
Did you know that it doesn't appear to do this in the JDBC driver?

PR #4, opened 2018-06-22
Initialize reason and SQLState in PostgresqlServerErrorException.

mp911de:
We now initialize reason and SQLState properties of R2dbcException
to provide details for code that uses only R2dbcException.
We're aligning the behavior with the JDBC driver that only
populates SQLState and leaves errorCode empty.

Closes gh-3.


mp911de:
Looks better.

PR #20, opened 2018-10-11
further reduction of Stream usage in Binding

davecramer:
I'm curious though about the changes I had to make in BindingTest?

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
@@ -27,7 +27,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-
+import java.util.Optional;
 
 
 import static io.r2dbc.postgresql.message.backend.Field.FieldType.CODE;
@@ -167,7 +167,7 @@ public final class PostgresqlServerErrorException extends R2dbcException {
      *
      * @return the value of the {@link FieldType#COLUMN_NAME} field
      */
-    public String getColumnName() { return this.columnName; }
+    public Optional<String> getColumnName() { return Optional.ofNullable(this.columnName); }
 
     /**
      * Returns the value of the {@link FieldType#CONSTRAINT_NAME} field.
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
@@ -23,11 +23,12 @@ import io.r2dbc.postgresql.message.backend.Field.FieldType;
 import io.r2dbc.spi.R2dbcException;
 import reactor.core.publisher.SynchronousSink;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
-import java.util.stream.Collectors;
+
+
 
 import static io.r2dbc.postgresql.message.backend.Field.FieldType.CODE;
 import static io.r2dbc.postgresql.message.backend.Field.FieldType.COLUMN_NAME;
@@ -166,17 +167,15 @@ public final class PostgresqlServerErrorException extends R2dbcException {
      *
      * @return the value of the {@link FieldType#COLUMN_NAME} field
      */
-    public Optional<String> getColumnName() {
-        return Optional.ofNullable(this.columnName);
-    }
+    public String getColumnName() { return this.columnName; }
 
     /**
      * Returns the value of the {@link FieldType#CONSTRAINT_NAME} field.
      *
      * @return the value of the {@link FieldType#CONSTRAINT_NAME} field
      */
-    public Optional<String> getConstraintName() {
-        return Optional.ofNullable(this.constraintName);
+    public String getConstraintName() {
+        return this.constraintName;
     }
 
     /**
@@ -184,8 +183,8 @@ public final class PostgresqlServerErrorException extends R2dbcException {
      *
      * @return the value of the {@link FieldType#DATA_TYPE_NAME} field
      */
-    public Optional<String> getDataTypeName() {
-        return Optional.ofNullable(this.dataTypeName);
+    public String getDataTypeName() {
+        return this.dataTypeName;
     }
 
     /**
@@ -193,8 +192,8 @@ public final class PostgresqlServerErrorException extends R2dbcException {
      *
      * @return the value of the {@link FieldType#DETAIL} field
      */
-    public Optional<String> getDetail() {
-        return Optional.ofNullable(this.detail);
+    public String getDetail() {
+        return this.detail;
     }
 
     /**
@@ -202,8 +201,8 @@ public final class PostgresqlServerErrorException extends R2dbcException {
      *
      * @return the value of the {@link FieldType#FILE} field
      */
-    public Optional<String> getFile() {
-        return Optional.ofNullable(this.file);
+    public String getFile() {
+        return this.file;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


mp911de:
Awesome, thanks a lot, Dave. Tests shouldn't require a change. If so, then we should find out why this is.

Care to stick to `List<…>` instead of `ArrayList<…>` in local variable declarations?

davecramer:
> Awesome, thanks a lot, Dave. Tests shouldn't require a change. If so, then we should find out why this is.

Ya, I thought that was strange too, but the test before made no sense when I looked at it.
> 
> Care to stick to `List<…>` instead of `ArrayList<…>` in local variable declarations?

sure..

src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
```diff
@@ -166,80 +167,78 @@ public String getCode() {
      *
      * @return the value of the {@link FieldType#COLUMN_NAME} field
      */
-    public Optional<String> getColumnName() {
```
nebhale: I'd prefer not to remove `Optional` in this error.  It's method signatures are accurate, it's not in a hot code path, and given how the PostgreSQL error is such a bag of mostly empty fields, I think pointing out the mostly emptiness is worthwhile. @mp911de thoughts?

src/main/java/io/r2dbc/postgresql/client/TransactionStatus.java
```diff
@@ -50,10 +49,10 @@
     static TransactionStatus valueOf(ReadyForQuery.TransactionStatus t) {
         Objects.requireNonNull(t, "t must not be null");
 
-        return Arrays.stream(values())
-            .filter(type -> type.discriminator == t)
-            .findFirst()
-            .orElseThrow(() -> new IllegalArgumentException(String.format("%s is not a valid transaction status", t)));
+        if ( t == IDLE.discriminator ) return IDLE;
```
nebhale: My kingdom for an enumeration-aware `switch` statement.  Can you please add curly braces here? I know it's more verbose, but we're an always-curly braces project.

src/main/java/io/r2dbc/postgresql/message/Format.java
```diff
@@ -49,10 +47,9 @@
      * @return the enum constant of this type with the specified discriminator
      */
     public static Format valueOf(short s) {
-        return Arrays.stream(values())
-            .filter(type -> type.discriminator == s)
-            .findFirst()
-            .orElseThrow(() -> new IllegalArgumentException(String.format("%d is not a valid format", s)));
+        if ( s == BINARY.discriminator ) return BINARY;
```
nebhale: Curly braces here as well please.

src/main/java/io/r2dbc/postgresql/client/TransactionStatus.java
nebhale: Actually, scratch the previous.  Can you please match this to the style you used in `BackendMessageDecoder` instead?

src/main/java/io/r2dbc/postgresql/message/Format.java
nebhale: Actually, scratch the previous. Can you please match this to the style you used in `BackendMessageDecoder` instead?

src/main/java/io/r2dbc/postgresql/message/backend/ReadyForQuery.java
```diff
@@ -104,10 +103,10 @@ static ReadyForQuery decode(ByteBuf in) {
         }
 
         static TransactionStatus valueOf(byte b) {
-            return Arrays.stream(values())
-                .filter(type -> type.discriminator == b)
-                .findFirst()
-                .orElseThrow(() -> new IllegalArgumentException(String.format("%c is not a valid transaction status", b)));
+            if ( b == IDLE.discriminator ) return IDLE;
```
nebhale: Can you please match this to the style you used in `BackendMessageDecoder` instead?

src/test/java/io/r2dbc/postgresql/client/BindingTest.java
```diff
@@ -52,7 +52,7 @@ void getParameterFormats() {
         binding.add(0, new Parameter(BINARY, 100, TEST.buffer(4).writeInt(200)));
         binding.add(2, new Parameter(BINARY, 100, TEST.buffer(4).writeInt(300)));
 
-        assertThat(binding.getParameterFormats()).containsExactly(BINARY, TEXT, BINARY);
```
nebhale: No, this shouldn't need to change and if these fail, it's a bug in the updated implementations.  What these are ensuring is that if you only bind parameters 0 and 2, then parameter 1 (which must exist) is contributed as an unspecified binding of some kind.  In your new implementation, you only iterate over the number of bindings that exist, potentially moving bindings from their specified index to an earlier one to fill in holes.

src/main/java/io/r2dbc/postgresql/message/Format.java
davecramer: given there are only 2 the loop seems redundant ? I'd be surprised if the compiler didn't optimize it out.

src/test/java/io/r2dbc/postgresql/client/BindingTest.java
davecramer: Damn, I missed the fact you were skipping one.

I would think that this would give surprising results if left this way.

I would expect this to create an error instead of silently adding a TEXT format in the middle.

src/main/java/io/r2dbc/postgresql/message/Format.java
nebhale: My main concern is consistency here.  I'd like people to look into any enumeration and see code that they're familiar with.  I'm with you that the compiler likely optimizes this out, and given that, I'd like to use the more generic form so that it stretches across all implementations.

src/test/java/io/r2dbc/postgresql/client/BindingTest.java
nebhale: Interesting.  That note in 51.2.3 seems to indicate that this is _possible_, but you think it's better that clients don't support the behavior.  I'm can definitely get on board with that if you'd like to update the implementation to behave that way.

davecramer: are you referring to 52.1.3 https://www.postgresql.org/docs/10/static/protocol-overview.html#PROTOCOL-FORMAT-CODES

If so I'm confused...

davecramer: OK, reading this again...
I understand the note in 51.2.3 to be saying. If you want the backend to infer the type of the parameter then set it to UNSPECIFIED (0). However by simply not binding anything as per the current test there is no Parameter.value. So yes, I would think this would be surprising to the user. Now if you want to  be able to have the server infer the type of the *value* then we should set the type to 0

On another note 51.2.3 refers to the 9.6 docs. I tend to use the latest which in a few days will be 11. I'm not sure how to deal with this, other than full href's

nebhale: Sounds good.  There are two options here then I think.  The first is to hold fire on this PR and start a second one to update existing behavior (pre-UNKNOWN) to throw an exception somewhere along the line if bindings are not contiguous.  The second option is to include that update as part of this change, which I'm OK with as the change without that (once it's been updated to encode-only) is quite small.  I'll leave it to you to decide how to go about it.

src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
mp911de: Exception use is fine. If an exception becomes the hot code path, then I guess we have a different kind of problem then.

src/test/java/io/r2dbc/postgresql/client/BindingTest.java
davecramer: OK, lets undo the changes to binding in this PR and commit what we agree on now, then fix the binding issue in another PR. History will be better served that way I believe.

davecramer:
I suggest we close this in favour of #24

nebhale:
Closing in favor of #24

PR #22, opened 2018-10-15
Add additional decoding for TEXT, CHAR, AND UNKNOWN

davecramer:
Removed UNSPECIFIED type, replace with UNKNOWN
Pretty sure there is no type or class in postgresql where the oid is 2
UNKNOWN is a real possibility though, and in my opinion a better default.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/client/Binding.java
+++ b/src/main/java/io/r2dbc/postgresql/client/Binding.java
@@ -28,14 +28,14 @@ import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import static io.r2dbc.postgresql.message.Format.TEXT;
-import static io.r2dbc.postgresql.type.PostgresqlObjectId.UNSPECIFIED;
+import static io.r2dbc.postgresql.type.PostgresqlObjectId.UNKNOWN;
 
 /**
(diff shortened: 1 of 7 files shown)


davecramer:
OK, I think unspecified is for incoming rows, not results, but I'm not sure either.
for UNKNOWN. If you do "select  'hello' as foo" foo will have a type of UNKNOWN in 9.6, in 10 it is String.
As for docs, none that I am aware of. 
It was used in situations where strings could be coerced to timestamps etc, 

above is one example.

The following throws an error currently

 @Test
    void getResultSet() throws  Exception{

        Flux <String>foo = Mono
            .from(this.connectionFactory
                .create())
            .flatMapMany(connection -> Flux.from(
                connection.createStatement("select 'hello' as foo").execute())
                .flatMap(result -> result.map((row, rowMetadata) -> row.get("foo", String.class))));
        foo.subscribe(System.out::println, Throwable::printStackTrace);
        Thread.sleep(3000);

    }

nebhale:
So given that `UNSPECIFIED` is for bindings, should this PR be updated to only make the change for decoding?

davecramer:
Yes, I will fix this..

davecramer:
close this in favour of #25

PR #28, opened 2018-10-21
Fix issue with simultaneous query execution for the same connection

OlegDokuka:
The following PR fix issue when we execute several Queries to the same connection.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,7 +41,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <r2dbc-spi.version>1.0.0.BUILD-SNAPSHOT</r2dbc-spi.version>
-        <reactor.version>Californium-RELEASE</reactor.version>
+        <reactor.version>Californium-SR1</reactor.version>
         <slf4j.version>1.7.25</slf4j.version>
         <spring-boot.version>2.0.5.RELEASE</spring-boot.version>
         <testcontainers.version>1.9.1</testcontainers.version>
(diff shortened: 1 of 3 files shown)


nebhale:
@OlegDokuka I'm not sure why this would need to be solved.  PostgreSQL doesn't allow multiplexing on connections and pipelining isn't dependable or desirable.  We'd planned on honoring the same semantics the database expects that there is only a single "user" of a connection at a given time.  This would imply that you wouldn't have "simultaneous" queries on a single connection.

OlegDokuka:
Hey @nebhale.

Take a look at the code. Either it is supported or not, one can write simultaneous execution and we do not have protection from that. Recently I did a demo for the presentation of r2dbc, and will not be able to run simple case during the transaction.
Since transaction uses the same connection, it is really easy to subscribe twice to the same source, since responseEmitter.next() does not guarantee that each new subscriber receives different element.

Thus, this PR makes sense, until there is another protection will be introduced

OlegDokuka:
It is possible that I misunderstood you, do you mean provide protection on the higher level?

nebhale:
Well, no protection per se, just an understanding that like JDBC or the database drivers in other languages, only a single query can be executed at a given time.  In a pooling scenario, you'd take a connection from a pool, execute your query and get your results and then return the connection to the pool for the next person to use it.

nebhale:
@OlegDokuka Not sure what you mean that for non-transaction it does not work.  If you follow a transaction to its completion (commit or rollback), you can safely hand back the connection for further usage.

OlegDokuka:
The problem is the following. You have a database client. Then you start a transaction and receive a connection. Then you execute two queries using zip. In the client, those queries will be sent to the same `ReactorNettyClient` over `#exchange` with the same connection. Both will be enqueued in the `requestEmitter` processor and then both will be subscribed to responseEmitter using `next` method. Both will receive the same UnicastWindowFluxInner and some of the will get an error. This is the problem.

nebhale:
Ah, yeah the use of `zip()`.  That makes sense.

mp911de:
I assume this fix is widely applicable for R2DBC drivers that use networking without an async-ready protocol.

nebhale:
Multiplexed protocol would be the key I think.  The ability to have more than one simultaneous query over a single socket connection.

PR #29, opened 2018-10-26
Ensure request and response subscribers complete when channel is unregistered.

chang-chao:
In some cases , such as https://github.com/r2dbc/r2dbc-postgresql/issues/17 , PostgreSQL server does not close the connection gracefully , where `responseProcessor` can't receive the `onComplete` signal. This causes the `Connection.close()` to never end.
This PR fixes this by ensuring connection subscribers to the end in case of channel deregistration.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
+++ b/src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
@@ -113,13 +113,15 @@ public final class ReactorNettyClient implements Client {
             sink.next(message);
         });
 
-    private class EnsureSubsribersCompleteChannelHandler extends ChannelDuplexHandler {
+    private class EnsureSubscribersCompleteChannelHandler extends ChannelDuplexHandler {
 
         @Override
         public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
             super.channelUnregistered(ctx);
             requestProcessor.onComplete();
-            responseProcessor.onComplete();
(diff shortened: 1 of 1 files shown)


src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
```diff
@@ -107,6 +109,16 @@
             sink.next(message);
         });
 
+    private class EnsureSubsribersCompleteChannelHandler extends ChannelDuplexHandler {
```
nebhale: Spelling error.

chang-chao: fixed

src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
```diff
@@ -111,6 +113,18 @@
             sink.next(message);
         });
 
+    private class EnsureSubscribersCompleteChannelHandler extends ChannelDuplexHandler {
+
+        @Override
+        public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
+            super.channelUnregistered(ctx);
+            requestProcessor.onComplete();
+            for (MonoSink<Flux<BackendMessage>> responseReceiver = responseReceivers.poll(); responseReceiver != null; ) {
```
nebhale: Note that this still needed `responseReceivers.poll()` as the third argument to the for loop to go to the next element in the `Queue`.  I'll fix it in a following commit.

chang-chao: Oh，forgot that，thanks！

PR #30, opened 2018-10-29
Handle authentication error response in creating connections.

chang-chao:
Before this PR, error responses in the StartupMessageFlow are ignored,  thus the connection mono can be completed even if the underlying connection failed.
This PR fixes this issue.


nebhale:
@OlegDokuka Another example of where a custom operator would have made it much easier to notice and solve this problem.

nebhale:
@OlegDokuka Another example of where a custom operator would have made it much easier to notice and solve this problem.

Issue #42, opened 2018-12-05
Postgres Arrays: Support for multi-dimensional arrays


mp911de:
Right now, the driver supports single-dimensional arrays. Postgres also supports multi-dimensional arrays such as `int[][]`:

```sql
CREATE TABLE tictactoe (
    squares   integer[3][3]
);
```

The driver should also support multi-dimensional arrays.

See also: https://www.postgresql.org/docs/11/arrays.html

/cc @rdegnan

nebhale:
resolved by #54

PR #53, opened 2019-01-10
Accept multidimensional array as parameter type in array codecs

fred84:
Implementation of multidimensional arrays (#42) are split into 2 parts:
- support multidimensional java array as type parameter and additional type validation (this PR)
- implementations of `doEncode`, `doEncodeItem`, `doDecode`, `doDecodeItem`

I think doing it as a single PR would be too big for review.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
@@ -28,25 +28,40 @@ import java.util.List;
 
 import static io.r2dbc.postgresql.message.Format.FORMAT_BINARY;
 
-abstract class AbstractArrayCodec<T> extends AbstractCodec<T[]> {
+abstract class AbstractArrayCodec<T> extends AbstractCodec<Object[]> {
 
     private static final byte DELIMITER = ',';
 
     private final ByteBufAllocator byteBufAllocator;
+    private final Class<T> componentType;
 
-    AbstractArrayCodec(ByteBufAllocator byteBufAllocator, Class<T[]> type) {
-        super(type);
+    AbstractArrayCodec(ByteBufAllocator byteBufAllocator, Class<T> componentType) {
+        super(Object[].class);
+        this.componentType = componentType;
         this.byteBufAllocator = Assert.requireNonNull(byteBufAllocator, "byteBufAllocator must not be null");
     }
 
     abstract T decodeItem(ByteBuf byteBuf, Format format, Class<?> type);
 
+    boolean isTypeAssignable(Class<?> type) {
+        if (!type.isArray()) {
+            return false;
+        }
+
+        return getBaseComponentType(type).equals(componentType);
+    }
+
     @Override
-    @SuppressWarnings("unchecked")
-    final T[] doDecode(ByteBuf byteBuf, Format format, Class<? extends T[]> type) {
+    public boolean canEncode(Object value) {
+        return isTypeAssignable(value.getClass());
+    }
+
+    @Override
+    final Object[] doDecode(ByteBuf byteBuf, Format format, Class<? extends Object[]> type) {
         Assert.requireNonNull(byteBuf, "byteBuf must not be null");
         Assert.requireNonNull(format, "format must not be null");
         Assert.requireNonNull(type, "type must not be null");
+        assertArrayDimension(type, 1);
 
         List<T> items = new ArrayList<>();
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
```diff
@@ -94,5 +111,40 @@ final Parameter doEncode(T[] value) {
...
+
+        Class<?> currentType = type;
+        while (currentType.isArray()) {
+            currentType = currentType.getComponentType();
+        }
+
+        return currentType;
+    }
 
+    private static int getDimensions(Class<?> type) {
```
nebhale: Can be simpler with a single loop that might never execute:

```java
Class<?> t = type;

int dimensions = 0;
while (t.isArray()) {
    t = t.getComponentType();
    dimensions++;
}

return dimensions;
```

src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
```diff
@@ -94,5 +111,40 @@ final Parameter doEncode(T[] value) {
 
     abstract void encodeItem(ByteBuf byteBuf, T value);
 
+    private static Class<?> getBaseComponentType(Class<?> type) {
```
nebhale: Can be simpler with a loop that might never execute:

```java
Class<?> t = type;

while (t.isArray()) {
    t = t.getComponentType();
}

return t;
```

src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
```diff
@@ -28,25 +28,40 @@
...
     }
 
     abstract T decodeItem(ByteBuf byteBuf, Format format, Class<?> type);
 
+    boolean isTypeAssignable(Class<?> type) {
+        if (!type.isArray()) {
+            return false;
+        }
+
+        return getBaseComponentType(type).equals(componentType);
```
nebhale: Should be `isAssignableFrom` rather than `equals`.

src/main/java/io/r2dbc/postgresql/codec/AbstractCodec.java
```diff
@@ -36,12 +36,16 @@ public final boolean canDecode(int dataType, Format format, Class<?> type) {
         Assert.requireNonNull(format, "format must not be null");
         Assert.requireNonNull(type, "type must not be null");
 
-        return type.isAssignableFrom(this.type) &&
+        return isTypeAssignable(type) &&
             doCanDecode(format, PostgresqlObjectId.valueOf(dataType));
     }
 
+    boolean isTypeAssignable(Class<?> type) {
+        return type.isAssignableFrom(this.type);
```
nebhale: Need to enforce nullability.

src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
```diff
@@ -28,25 +28,40 @@
...
+    AbstractArrayCodec(ByteBufAllocator byteBufAllocator, Class<T> componentType) {
+        super(Object[].class);
+        this.componentType = componentType;
         this.byteBufAllocator = Assert.requireNonNull(byteBufAllocator, "byteBufAllocator must not be null");
     }
 
     abstract T decodeItem(ByteBuf byteBuf, Format format, Class<?> type);
 
+    boolean isTypeAssignable(Class<?> type) {
+        if (!type.isArray()) {
```
nebhale: Need to enforce nullability.

src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
```diff
@@ -28,25 +28,40 @@
...
+        }
+
+        return getBaseComponentType(type).equals(componentType);
+    }
+
     @Override
-    @SuppressWarnings("unchecked")
-    final T[] doDecode(ByteBuf byteBuf, Format format, Class<? extends T[]> type) {
+    public boolean canEncode(Object value) {
+        return isTypeAssignable(value.getClass());
```
nebhale: Need to enforce nullability.

nebhale:
@fred84 Why is it important to assert the size of the array in `doDecode`?  Doesn't the `Class<? extends Object[]> type` parameter declaration already ensure this?

src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
```diff
@@ -94,5 +111,40 @@ final Parameter doEncode(T[] value) {
...
+        while (currentType.isArray()) {
+            currentType = currentType.getComponentType();
+            dimensions++;
+        }
+
+        return dimensions;
+    }
+
+    private static void assertArrayDimension(Class<?> type, int allowed) {
+        int actual = getDimensions(type);
```
nebhale: Should be part of the `Assert` class.

src/test/java/io/r2dbc/postgresql/codec/IntegerArrayCodecTest.java
```diff
@@ -105,4 +106,13 @@ void encodeNull() {
             .isEqualTo(new Parameter(FORMAT_TEXT, INT4_ARRAY.getObjectId(), null));
     }
 
+    @Test
+    void multidimensionalArrayNotYetAllowed() {
+        IntegerArrayCodec codec = new IntegerArrayCodec(TEST);
+
+        assertThrows(
```
nebhale: Use the `assertIllegalArgumentException()` convenience method.

src/test/java/io/r2dbc/postgresql/codec/IntegerArrayCodecTest.java
```diff
@@ -105,4 +106,13 @@ void encodeNull() {
             .isEqualTo(new Parameter(FORMAT_TEXT, INT4_ARRAY.getObjectId(), null));
     }
 
+    @Test
+    void multidimensionalArrayNotYetAllowed() {
```
nebhale: All array codecs need this test.

fred84:
@nebhale
1) canDecode only checks that expected type is array and that is base component type matches codec. Number of dimensions is ignore there.
2) multidimesional array is object array. E.g. `new int[][] {{1}} instanceof Object[]` is true.  Until next PR we only support 1-dimensional arrays, so this assert will reject arrays with more dimensions.
3) in next PR I'll match number of dimensions in expected type with number of dimensions from postgresql

nebhale:
Sounds good.  `master` should be ready for the next bit.

PR #54, opened 2019-01-14
postgresql multidimensional array implementation

fred84:
Implementation for #42 .

Notes:
- binary format is not tested with real queries. I took value dumps from queries with current postgresql jdbc driver and placed them inside unit tests. We should consider implementing "force binary" option to be able to write integration tests for binary format;
- IntegerArrayCodecTest has much more tests then Long/Short/String* siblings. I think we should parametrize these test to avoid copy-paste (not included in current PR);
- AbstractArrayCodec::buildArrayList is very similar to original https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jdbc/PgArray.java#L413 implementation


fred84:
Support for decoding array-type into Object (#69) added to this PR.

Issue #67, opened 2019-02-01
Row.get(…) fails for array-typed values


mp911de:
Calling `Row.get(…)` or `Row.get(…, Object.class)` fails for array-typed columns with:

```
java.lang.IllegalArgumentException: Cannot decode value of type java.lang.Object
```

## Expectation

`Row.get(…)` should return the array value with the most appropriate type. For `INT[]` that should be `Integer[].class`.

## Steps to reproduce

SQL:

```sql
CREATE TABLE with_arrays (my_array INT[]);
INSERT INTO with_arrays (my_array) VALUES('{1,2,3}')
```

Code:

```java
Mono.from(connectionFactory.create())
    .flatMapMany(it -> {
        return it.createStatement("SELECT my_array FROM with_arrays").execute();
    }).flatMap(it -> it.map((row, rowMetadata) -> row.get("my_array")))
    .doOnNext(System.out::println)
    .as(StepVerifier::create)
    .expectNextCount(1)
    .verifyComplete();
```

nebhale:
This may be because it’s [simply incomplete](
https://github.com/r2dbc/r2dbc-postgresql/pull/54).  Guess that one _does_ need to make it in for M7.

mp911de:
There was some change between M6 and M7 as it has worked with M6.

Squiry:
Well, I can fix this with a few lines, but the whole `Object.class` decoding (not only for arrays) thing stands on type erasure, I'm not sure how bad is that.

nebhale:
Resolved by #69

PR #69, opened 2019-02-01
Fix for decoding array-typed value into Object

Squiry:
[#67]

> Well, I can fix this with a few lines, but the whole Object.class decoding (not only for arrays) thing stands on type erasure, I'm not sure how bad is that.


mp911de:
Thanks a lot for your pull request. We have a PR #54 that introduces multi-dimensional arrays and this pull request duplicates some of the changes in #54. I propose to drop this one in favor of #54. What do you think?

Squiry:
#54 fixed only half of the problem (`AbstractArrayCodec` part): `DefaultCodecsTest.delegatePriority` is still failing with `Object.class` argument and array data. We need to fix AbstractCodec anyway (@fred84 can do that in his pull request ofc).

fred84:
@Squiry I'll include this change into my PR

Issue #34, opened 2018-11-03
Error message for authentication failure is not very useful


davecramer:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Background

PostgreSQL returns a single [`ErrorResponse`](https://www.postgresql.org/docs/11/protocol-message-formats.html) message.  When an authentication error is returned it's not clear that this common error has occurred.

## Problem

Right now, it's not obvious when an authentication error, which is very common, has occurred.

## Solution

An `R2dbcAuthenticationException` should be created as a subclass of [`R2dbcException`](https://github.com/r2dbc/r2dbc-spi/blob/master/r2dbc-spi/src/main/java/io/r2dbc/spi/R2dbcException.java) to indicate that an authentication error has occurred.  This type should then be returned when the [`ReactotNettyClient`](https://github.com/r2dbc/r2dbc-postgresql/blob/master/src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java#L72-L76) encounters a `28P01` error code message.

## Steps to fix

- [ ] Claim this issue with a comment below and ask any clarifying questions you need.
- [ ] Try to fix the issue following the steps above.
- [ ] Commit your changes and start a pull request.

## Deliverables
- [ ] `R2dbcAuthenticationException` in [`r2dbc-spi`](https://github.com/r2dbc/r2dbc-spi)
- [ ] Updated [`ConnectionFactory`](https://github.com/r2dbc/r2dbc-spi/blob/master/r2dbc-spi/src/main/java/io/r2dbc/spi/ConnectionFactory.java#L31) Javadoc indicating that this error can be emitted.
- [ ] Updated [`ReactorNettyClient`](https://github.com/r2dbc/r2dbc-postgresql/blob/master/src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java) that maps `28P01` errors to an `R2dbcAuthenticationException`.

davecramer:
It's possible that I over reacted to this after reading the whole log message.

I'm thinking that a top-level exception would be less surprising though.

nebhale:
[...]

PR #72, opened 2019-02-07
Feature/add auth exception

uaihebert:
Check for the error code and see if it is related with Auth authentication.

The idea of this PR is to give a clear error message. It will be displayed like:
image
Of course, this is the first idea, feedbacks are welcome.

Relates to issue: #34

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.r2dbc.postgresql;
+
+import io.r2dbc.postgresql.message.backend.Field;
+
+import java.util.List;
+
+/**
+ * An exception that represents a authentication error.
+ */
+public class PostgresqlAuthenticationFailure extends PostgresqlServerErrorException {
+
+    /**
+     * Creates a new exception.
+     *
+     * @param fields the fields to be used to populate the exception
+     * @throws IllegalArgumentException if {@code fields} is {@code null}
+     */
+    public PostgresqlAuthenticationFailure(final List<Field> fields) {
+        super(fields);
+    }
+
+    @Override
+    public String toString() {
+        return "PostgresqlAuthenticationFailure: " + getMessage();
+    }
+}
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
@@ -31,7 +31,7 @@ public final class PostgresqlAuthenticationFailure extends PostgresqlServerError
      * @param fields the fields to be used to populate the exception
      * @throws IllegalArgumentException if {@code fields} is {@code null}
      */
-    public PostgresqlAuthenticationFailure(final List<Field> fields) {
+    public PostgresqlAuthenticationFailure(List<Field> fields) {
         super(fields);
     }
 
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
@@ -23,7 +23,7 @@ import java.util.List;
 /**
  * An exception that represents a authentication error.
  */
-public class PostgresqlAuthenticationFailure extends PostgresqlServerErrorException {
+public final class PostgresqlAuthenticationFailure extends PostgresqlServerErrorException {
 
     /**
      * Creates a new exception.
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.r2dbc.postgresql.exception;
+package io.r2dbc.postgresql;
 
 import io.r2dbc.postgresql.message.backend.ErrorResponse;
 import io.r2dbc.postgresql.message.backend.Field;
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
@@ -52,7 +52,7 @@ import static io.r2dbc.postgresql.message.backend.Field.FieldType.WHERE;
 /**
  * An exception that represents a server error.  This exception is a direct translation of the {@link ErrorResponse} message.
  */
-public final class PostgresqlServerErrorException extends R2dbcException {
+public class PostgresqlServerErrorException extends R2dbcException {
 
     private final String code;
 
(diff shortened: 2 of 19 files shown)


pivotal-issuemaster:
@uaihebert Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=r2dbc/r2dbc-postgresql&pullRequestId=72)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=r2dbc/r2dbc-postgresql&pullRequestId=72) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@uaihebert Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=r2dbc/r2dbc-postgresql&pullRequestId=72)!

src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
```
mp911de: Nit: License header date reflects the inception of a file. So in this case `2019` is sufficient.

src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
```diff
@@ -0,0 +1,42 @@
...
+package io.r2dbc.postgresql;
+
+import io.r2dbc.postgresql.message.backend.Field;
+
+import java.util.List;
+
+/**
+ * An exception that represents a authentication error.
+ */
+public class PostgresqlAuthenticationFailure extends PostgresqlServerErrorException {
```
mp911de: We encapsulate and close classes as much as possible. This class should be `final`.

src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Creates a new exception.
+     *
+     * @param fields the fields to be used to populate the exception
+     * @throws IllegalArgumentException if {@code fields} is {@code null}
+     */
+    public PostgresqlAuthenticationFailure(final List<Field> fields) {
```
mp911de: No need to make method args `final`.

src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
```diff
@@ -347,11 +348,27 @@ public String toString() {
...
     static void handleErrorResponse(BackendMessage message, SynchronousSink<BackendMessage> sink) {
-        if (message instanceof ErrorResponse) {
-            sink.error(PostgresqlServerErrorException.toException((ErrorResponse) message));
-        } else {
+        if (!(message instanceof ErrorResponse)) {
             sink.next(message);
+            return;
         }
+
+        final ErrorResponse error = (ErrorResponse) message;
```
mp911de: No need to make variables `final`. This causes mostly noise while reading code.

src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
```diff
@@ -347,11 +348,27 @@ public String toString() {
     }
 
     static void handleErrorResponse(BackendMessage message, SynchronousSink<BackendMessage> sink) {
```
mp911de: Probably we want to move that code into an `ExceptionFactory`, wdyt @nebhale? 

The issue here is that we're creating a cycle between `PostgresqlServerErrorException ` and `PostgresqlAuthenticationFailure`. I think we can take care of this during the merge.

src/test/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailureTest.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
```
mp911de: Nit: License header date reflects the inception of a file. So in this case `2019` is sufficient.

src/test/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailureTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static io.r2dbc.postgresql.PostgresqlConnectionFactoryProvider.POSTGRESQL_DRIVER;
+import static io.r2dbc.spi.ConnectionFactoryOptions.DATABASE;
+import static io.r2dbc.spi.ConnectionFactoryOptions.DRIVER;
+import static io.r2dbc.spi.ConnectionFactoryOptions.HOST;
+import static io.r2dbc.spi.ConnectionFactoryOptions.PASSWORD;
+import static io.r2dbc.spi.ConnectionFactoryOptions.PORT;
+import static io.r2dbc.spi.ConnectionFactoryOptions.USER;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+public class PostgresqlAuthenticationFailureTest {
```
mp911de: Nit: JUnit 5 tests classes can be reduced to package-default visibility. Our test classes are typically `final`.

src/test/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailureTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @RegisterExtension
+    static final PostgresqlServerExtension SERVER = new PostgresqlServerExtension();
+
+    @Test
+    public void authExceptionCausedByWrongCredentials() {
```
mp911de: Nit: Test method visibility can be reduced to package-default visibility.

src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
uaihebert: Hey, I just copied over this header from other files.

I will update this date.

src/main/java/io/r2dbc/postgresql/PostgresqlAuthenticationFailure.java
uaihebert: ok. I just got used to my company checkstyle 😂

src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
uaihebert: I could create a Factory, it is not a problem.

I thought about adding one, but I tried to touch the minimum as possible. I did not want to update too much code.

uaihebert:
hello @mp911de 

I have pushed the changes.

regarding the Unit tests, I added the same as: `PostgresqlServerErrorExceptionTest`.

The deal is, I do not like repeating code. :D

I think we could create a "generic exception Test" that would test both of exceptions, since they all have the same kind of tests. 

it could be a separated issue to avoid making this too big.

src/main/java/io/r2dbc/postgresql/PostgresqlServerErrorException.java
mp911de: In general, that's a good guideline to touch as little as necessary. In this case, it would make sense to go the additional step in changing how exceptions are created.

uaihebert: Hello, I just pushed the update with the factory.

mp911de:
Thanks a lot. Let me have a look today or tomorrow.

uaihebert:
Awesome. the idea was:

1. to create a factory that will manage the creation of the exception
2. to move out from the old exception the creation logic
3. to create a package for the exceptions, now there are 3 classes related to exceptions. this will "hide" the logic of exceptions inside a default scope

thank you for the support.

src/main/java/io/r2dbc/postgresql/exception/PostgresqlServerErrorException.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.r2dbc.postgresql;
+package io.r2dbc.postgresql.exception;
```
mp911de: What's the reason behind moving the exception to a separate package?

Typically, exceptions stay in the same package as the API itself to avoid package cycles and to keep exceptions on the same level as the code that emits the exception.

uaihebert: my idea here was to have the exceptions inside the exception related code. just as an organization of packages.

I thought it is confusing if I look into a package and I see stuff like:
- service.class
- exception.class
- factory.class

but, if this is not a problem for this project, I could move the exception back. ( :

mp911de: Separating domain types by technicality is in fact considered an anti-pattern because it distributes things that belong together. Having all domain functionality, that belongs together in a single place keeps code more cohesive and it is easier to identify things that belong together. 

A nice side effect of keeping types in the same package allows hiding internal classes (package private visibility).

uaihebert: I got your point. 

I just moved everything out of the package and delete it. 

Before merge, I will squash all commits.

Thank you for the explanation.

mp911de:
Given the upcoming changes in r2dbc/r2dbc-spi#47, I propose to wait until r2dbc/r2dbc-spi#53 is merged and then adapt the specific permission exception here.

Issue #21, opened 2018-10-14
implement binary transfer and codecs


davecramer:
Implement binary protocol
All of the codecs are TEXT only

nebhale:
Well, the [numerics are binary](https://github.com/r2dbc/r2dbc-postgresql/blob/master/src/main/java/io/r2dbc/postgresql/codec/IntegerCodec.java#L53-L71), but I take your point on the others.  Is this a bit of expertise you can lend?

davecramer:
yes, next on my list..
Also we need to tell the backend which oids we want in binary in the startup message.

nebhale:
Oh, that's interesting.  One thing I'd request, just since I don't have a deep background in PostgreSQL is that either commit messages or PRs include pointers to documentation for features like that.  I wasn't aware you could even do that :)

davecramer:
Actually I need to recant, it is sent in the prepare message to tell pg we can receive it in binary.

But sure, I'll try to document

davecramer:
I think  https://github.com/r2dbc/r2dbc-postgresql/blob/03b53d3e802f7700bf3bcff3f62131e0e2e31906/src/main/java/io/r2dbc/postgresql/client/ExtendedQueryMessageFlow.java#L103

needs to be changed so that Describe is before the bind so that we have the RowDescriptions to be able to send the binary formats to the bind command

mp911de:
See also #76.

PR #76, opened 2019-02-14
force binary implementation

fred84:
Implementation for #21 . According to [Message Formats](https://www.postgresql.org/docs/current/protocol-message-formats.html) we can specify result for for all columns.

> The number of result-column format codes that follow (denoted R below). This can be zero to indicate that there are no result columns or that the result columns should all use the default format (text); or one, in which case the specified format code is applied to all result columns (if any); or it can equal the actual number of result columns of the query.

If we can proceed with this solution, I'll add binary supports to rest of the codecs.


fred84:
Support for binary format is implemented for most codecs. Exceptions:
1) support for array codecs is implemented in #54 
2) binary format for date codecs doesn't support timezones and non-integer date format yet.

mp911de:
Thanks a lot for your pull request. `master` is now stable. Care to drop the workaround in `PostgresqlConnectionFactoryProvider `? Also, there are changes to the Maven wrapper that do not seem related to this change.

davecramer:
@fred84 I would not waste too much time with non-integer dates. I doubt anyone uses them anymore. The feature existed in a time where 64bit integers were not native. No package manager would build with floating point times
timezones are a challenge as postgres doesn't send them in binary 
Now for the fun one... Postgres supports a time of 24:00:00 see https://github.com/pgjdbc/pgjdbc/pull/1394 for some background

fred84:
@davecramer We could send ```SET TIMEZONE TO 'UTC'``` query after establishing connection and then treat all binary timestamps as in UTC zone. What do you think?

davecramer:
@fred84 the JDBC driver sets it to the client timezone. We tend to get it half right.. Half of the people like that, half don't. Pick one and document it.

You will also annoy people who actually issue 'SET TIMEZONE ...' Welcome to the wonderful world of driver writing

fred84:
@davecramer R2DBC currently supports binary format for backend messages (and this PR deal only with backend message decoding). Binary timestamps from backend are always in UTC (because they are stored in UTC in Postgres). So we are safe to treat them as UTC and do not need to know timezone of current connection. Am I right?

P.S. It is possible to set connection timezone in StartupMessage and we may need it in future

davecramer:
@fred84 well the first part of that statement is correct. As far as being safe to treat them as UTC I wish you good fortune. Time and timezones are the bane of all driver implementers.

Timestamps without timezones obviously have no timezone. Timestamps *WITH* timezone are still sent without timezones in binary. How you deal with this is up to you. As I said the JDBC driver sets the current timezone for the session to the clients TZ. We annoy some people and get it right for most. We do not use binary for timestamps. It's really not that huge a deal in reality.

fred84:
@davecramer Now connection timezone is set to system timezone.

nebhale:
@fred84 As part of merging this, I enabled three of the four tests you disabled.  The one I could do was the string array one and my read is that you didn't make the necessary changes to that codec (easy to forget given the size of the change).  If you'd like to submit another PR that deals with that last little bit, it'd be appreciated.

PR #96, opened 2019-04-07
Remove uneeded dependencies

sdeleuze:
As a low level library, I think `r2dbc-postgresql` should limit its dependencies. It seems Netty http support is not needed and `reactor-extra` is only used for tiny optional utility functions. I have chosen in this PR to remove usage of `function()` which is my preferred solution, another possibility is to copy `TupleUtils` in `r2dbc-postgresql`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -91,6 +91,16 @@
         <dependency>
             <groupId>io.projectreactor.netty</groupId>
             <artifactId>reactor-netty</artifactId>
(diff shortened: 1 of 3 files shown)


gregturn:
Being a part of the Reactor portfolio I’m not sure I see this as an extraneous dependency. Sure it’s synta sugar but it provides a consistent feature.

PR #104, opened 2019-05-01
TLS connectivity

### LINKED ISSUE #59: Investigate TLS connectivity

Squiry:
Resolves #59


src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -150,6 +166,10 @@ boolean isForceBinary() {
 
         private String schema;
 
+        private boolean ssl = false;
```
mp911de: We could probably add more methods to configure SSL certificates, but I think for now we might just start with default SSL properties and add all the sugar later.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -72,6 +79,16 @@ public PostgresqlConnectionFactory create(ConnectionFactoryOptions connectionFac
             builder.port(port);
         }
 
+        Boolean ssl = connectionFactoryOptions.getValue(SSL);
```
mp911de: 👍

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -72,6 +79,16 @@ public PostgresqlConnectionFactory create(ConnectionFactoryOptions connectionFac
...
+        if (ssl != null) {
+            builder.ssl(ssl);
+        }
+
+        SslProvider sslProvider = connectionFactoryOptions.getValue(SSL_PROVIDER);
```
mp911de: PostgresConnectionFactoryProvider is typically used through URL parsing. Let's omit `SslProvider` for now and add later something for certificate and verification mode properties.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -249,6 +270,29 @@ public Builder schema(@Nullable String schema) {
...
+        }
+
+
+        /**
+         * Configure {@link SslProvider} to use.
+         *
+         * @param sslProvider {@link SslProvider} to use
+         * @return this {@link Builder}
+         */
+        public Builder sslProvider(SslProvider sslProvider) {
```
mp911de: We should provide a method that accepts `io.netty.handler.ssl.SslContextBuilder`, too. Providing a `SslContextBuilder` requires fewer configuration objects on the usage side.

src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
```diff
@@ -110,12 +111,23 @@
...
-     * @param connection the TCP connection
+     * @param connection  the TCP connection
+     * @param sslProvider ssl provider for ssl protected connection
      * @throws IllegalArgumentException if {@code connection} is {@code null}
      */
-    private ReactorNettyClient(Connection connection) {
+    private ReactorNettyClient(Connection connection, @Nullable SslProvider sslProvider) {
         Assert.requireNonNull(connection, "Connection must not be null");
 
+        final Mono<Void> sslHandshakeFuture;
```
mp911de: Nit: No need for `final`.

src/main/java/io/r2dbc/postgresql/client/SSLSessionHandlerAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+        char response = (char) buf.readByte();
+        switch (response) {
+            case 'S':
+                ctx.channel().pipeline()
+                    .addFirst(this.sslHandler)
+                    .remove(this);
+                ctx.fireChannelRead(msg);
+                break;
+            case 'N':
+                throw new IllegalStateException("Server doesn't support SSL connections");
```
mp911de: Do we need throwing exceptions? Why don't we complete `handshakeFuture` exceptionally?

src/main/java/io/r2dbc/postgresql/client/SSLSessionHandlerAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.ssl.SslHandler;
+import io.r2dbc.postgresql.message.frontend.SSLRequest;
+import reactor.core.publisher.Mono;
+import reactor.netty.tcp.SslProvider;
+
+import java.util.concurrent.CompletableFuture;
+
+public class SSLSessionHandlerAdapter extends ChannelInboundHandlerAdapter {
```
mp911de: We have a similar setup in R2DBC SQL Server.

src/test/java/io/r2dbc/postgresql/client/ReactorNettyClientTest.java
```diff
@@ -205,7 +205,141 @@ private PostgresqlConnectionFactory createConnectionFactory(String username, Str
...
+        @Test
+        void exchangeSslWithPassword() throws IOException {
+            SslProvider sslProvider = SslProvider.builder()
+                .sslContext(SslContextBuilder.forClient()
+                    .trustManager(SERVER.getServerCrt().getInputStream())
+                )
+                .defaultConfiguration(TCP)
+                .build();
+
+            this.client = ReactorNettyClient.connect(SERVER.getHost(), SERVER.getPort(), sslProvider, null)
```
mp911de: Having all these `.block()` methods in the code is an anti-pattern as they don't time out. We should use at least `StepVerifier` where possible.

src/test/resources/ssl/client.crt
```diff
@@ -0,0 +1,30 @@
+-----BEGIN CERTIFICATE-----
```
mp911de: Please replace certificates with openssl certificate generation. These things are typically subject to expire and then no one knows how to set them up properly because certificates and keys aren't the most readable things. 

Take a look at Spring Vault's [`create_certificates.sh`](https://github.com/spring-projects/spring-vault/blob/master/src/test/bash/create_certificates.sh) that creates certificates as part of its build. We assume openssl availability but that should be less of an issue.

src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
```diff
@@ -195,9 +212,10 @@ private ReactorNettyClient(Connection connection) {
      * @param host               the host to connect to
      * @param port               the port to connect to
      * @param connectTimeout     connect timeout
+     * @param sslProvider        ssl provider for ssl protected connection
      * @throws IllegalArgumentException if {@code host} is {@code null}
      */
-    public static Mono<ReactorNettyClient> connect(ConnectionProvider connectionProvider, String host, int port, @Nullable Duration connectTimeout) {
+    public static Mono<ReactorNettyClient> connect(ConnectionProvider connectionProvider, String host, int port, @Nullable SslProvider sslProvider, @Nullable Duration connectTimeout) {
```
mp911de: Please encapsulate SSL config in a value object. Right now, the presence/absence of `SslProvider` makes some assumptions over SSL usage: Present: Require SSL. Absent SSL not supported. There are a lot of other options (see https://github.com/pgjdbc/pgjdbc/blob/master/docs/documentation/head/ssl-client.md) that we could want to support. And TBCH, having a non-nullable parameter is the better way to go.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -127,6 +135,14 @@ boolean isForceBinary() {
         return this.forceBinary;
     }
 
+    boolean getSsl() {
```
mp911de: We should return a value object here that encapsulates SSL usage. See comment below on `ReactorNettyClient.connect(…)`.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -249,6 +270,29 @@ public Builder schema(@Nullable String schema) {
             return this;
         }
 
+        /**
+         * Configure ssl usage.
+         *
+         * @param ssl
+         * @return this {@link Builder}
+         */
+        public Builder ssl(boolean ssl) {
```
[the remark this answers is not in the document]
Squiry: I can't see why we should use them at all. verify/verify-ca is a part of SSLContext and allow/require is for server to deside. I believe we should provide only SslProvider/SslProviderBuilder parameter and let user configure it.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
Squiry: No quite sure how to use this ConnectionFactoryOptions.SSL property at all, tbh.

src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
Squiry: If user configured SslContext then ssl requirement is a pretty good assumption. And other options are about configuring sslcontext. We should provide either them (with all that cert, key, password, etc params) and build sslcontext on our own or SslContext param.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
mp911de: `ConnectionFactoryOptions.SSL` is an indicator to enable SSL usage which means that we want to use SSL with full validation (CA, hostname, certificate validity) and fail if it's not supported/available.

src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
mp911de: Configuring `SslContext` isn't the most trivial thing to do. The API should support users in their intent to configure SSL in a convenient way.

atoko:
Most cloud providers enforce SSL these days, this would be greatly appreciated :)

maslick:
Any updates on TLS? I'm using postgres on Heroku, which requires SSL. Could you perhaps give an estimate when we can see this feature? Thanks a lot :)

mp911de:
Sorry for the long radio silence. We had to sort out a couple of issues across our team about how we can proceed with the project. We will need to fix SPI changes with #140 first before we can rebase and continue with this PR.

mp911de:
We have all prerequisites in place now. Care to rebase your work @Squiry so we can proceed with merging this PR?

Squiry:
@atoko just like in jdbc verification is enabled only in verify-ca and verify-full modes.
@mp911de sure, rebased.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -53,19 +53,42 @@
 
     private final String username;
 
-    private PostgresqlConnectionConfiguration(String applicationName, @Nullable Duration connectTimeout, @Nullable String database, boolean forceBinary, String host,
-                                              @Nullable Map<String, String> options, String password, int port, @Nullable String schema, String username) {
+    private final boolean ssl;
+
+    private final String sslCert;
```
mp911de: Nit: SSL fields except for `sslMode` should be `@Nullable`.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -162,6 +223,20 @@ boolean isForceBinary() {
 
         private String schema;
 
+        private boolean ssl = false;
+
+        private String sslCert = null;
```
mp911de: Nit: Fields initialized with `null` should be `@Nullable`.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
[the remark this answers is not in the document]
mp911de: Having `SSLMode` makes this property probably redundant. Alternatively, we could call `sslMode(REQUIRE)` here.

We can derive whether to use or to not use SSL from the provided `SslMode`.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -285,6 +360,83 @@ public Builder schema(@Nullable String schema) {
...
+            return this;
+        }
+
+        /**
+         * Configure ssl key.
+         *
+         * @param sslKey
+         * @return this {@link Builder}
+         */
+        public Builder sslKey(@Nullable String sslKey) {
```
mp911de: Nit: Let's document that this should be a path to the SSL key in PEM format.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -285,6 +360,83 @@ public Builder schema(@Nullable String schema) {
...
+            return this;
+        }
+
+        /**
+         * Configure ssl cert.
+         *
+         * @param sslCert
+         * @return this {@link Builder}
+         */
+        public Builder sslCert(@Nullable String sslCert) {
```
mp911de: Nit: Let's document that this should be a path to the SSL certificate in PEM format.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -285,6 +360,83 @@ public Builder schema(@Nullable String schema) {
...
+            return this;
+        }
+
+        /**
+         * Configure ssl password.
+         *
+         * @param sslPassword
+         * @return this {@link Builder}
+         */
+        public Builder sslPassword(@Nullable String sslPassword) {
```
mp911de: Nit: Let's switch to `Charsequence` to allow for implementations that do not enforce `String` caching in the JVM.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -285,6 +360,83 @@ public Builder schema(@Nullable String schema) {
...
+            return this;
+        }
+
+        /**
+         * Configure ssl root cert.
+         *
+         * @param sslRootCert
+         * @return this {@link Builder}
+         */
+        public Builder sslRootCert(@Nullable String sslRootCert) {
```
mp911de: Nit: Let's document that this should be a path to the SSL root/CA certificate in PEM format.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -186,7 +261,8 @@ public Builder applicationName(String applicationName) {
          */
         public PostgresqlConnectionConfiguration build() {
             return new PostgresqlConnectionConfiguration(this.applicationName, this.connectTimeout, this.database, this.forceBinary, this.host, this.options, this.password, this.port, this.schema,
-                this.username);
+                this.username,
```
mp911de: It would make sense to check whether the configured SSL key/cert/root cert actually exist to validate and fail early if the files are missing.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -89,6 +120,41 @@ public PostgresqlConnectionFactory create(ConnectionFactoryOptions connectionFac
             builder.options(options);
         }
 
+        Boolean ssl = connectionFactoryOptions.getValue(SSL);
+        if (ssl != null) {
+            builder.ssl(ssl);
+        }
+
+        SSLMode sslMode = connectionFactoryOptions.getValue(SSL_MODE);
```
mp911de: Typically, these values come from the URL parser and the URL parser has no knowledge over option types. So expect the value of `SSL_MODE` to potentially be a `String`.

src/main/java/io/r2dbc/postgresql/SSLConfig.java
```diff
@@ -0,0 +1,108 @@
...
+import reactor.netty.tcp.SslProvider;
+import reactor.util.annotation.Nullable;
+
+import javax.net.ssl.HostnameVerifier;
+import java.io.File;
+import java.util.function.Supplier;
+
+import static reactor.netty.tcp.SslProvider.DefaultConfigurationType.TCP;
+
+public class SSLConfig {
```
mp911de: Let's move that class into the `client` package to avoid package cycles and make it `final`.

src/main/java/io/r2dbc/postgresql/PGHostnameVerifier.java
```diff
@@ -0,0 +1,258 @@
...
+import java.net.IDN;
+import java.security.cert.CertificateParsingException;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+public class PGHostnameVerifier implements HostnameVerifier {
```
mp911de: Let's move that class into the `client` package to avoid package cycles and let's make it `final`.

src/main/java/io/r2dbc/postgresql/SSLMode.java
```diff
@@ -0,0 +1,76 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.r2dbc.postgresql;
+
+public enum SSLMode {
```
mp911de: Let's move this enum to the `client` package.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
Squiry: I agree that ssl parameter is redundant, I've just followed pgjdbc configurations like you said before and it has both.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
Squiry: Other fields has @Nullable only on their getters and constructor arguments, should I mark them nullable as well?

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
Squiry: SslContextBuilder.keyManager(File, File, String) requires String anyway, so we'll have to convert it to string anyway. Or should I make it CharSequence only in Builder fields?

src/main/java/io/r2dbc/postgresql/SSLConfig.java
Squiry: The configuration class getters has package-private access, we should change it to public then.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
mp911de: The idea is to help improve the status quo and postpone conversion to String as late as possible. Let’s keep passwords As CharSequence for as long as possible. I’ll take a look at the login password handling (not related to this PR) in the context of a new ticket.

src/main/java/io/r2dbc/postgresql/SSLConfig.java
mp911de: What class outside of the client package is calling the getters?

Squiry: SSLConfig in static factory method.

mp911de: What if we move `createSslConfig` into `PostgresqlConnectionConfiguration` so the configuration class can create  `SslConfig`?

PR #110, opened 2019-05-17
Blob encoding fix

Squiry:
Fix for a blob encoding issue.


mp911de:

PR #111, opened 2019-05-19
Fix for multiple usage of same parameter in extended query

Squiry:
Since c9bb24aa we can't use the same parameter twice (like 'SELECT $1 + $1').


mp911de:
Thanks a lot, that's merged now.

Issue #107, opened 2019-05-09
Weird codec issue when running concurrent queries


flyaruu:
I'm building a demo, and this issue seems to occur only when I run quite a few queries simultaneously, even though it seems too consistent to be race condition. Somehow different queries seem to interfere with each other's buffer.

This is the exception I get:
```
java.lang.NumberFormatException: For input string: " M"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:638)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at io.r2dbc.postgresql.codec.IntegerCodec.doDecode(IntegerCodec.java:61)
	at io.r2dbc.postgresql.codec.IntegerCodec.doDecode(IntegerCodec.java:32)
	at io.r2dbc.postgresql.codec.AbstractCodec.decode(AbstractCodec.java:67)
	at io.r2dbc.postgresql.codec.DefaultCodecs.decode(DefaultCodecs.java:93)
	at io.r2dbc.postgresql.PostgresqlRow.get(PostgresqlRow.java:83)
	at TestIssue.lambda$2(TestIssue.java:37)
	at io.r2dbc.postgresql.PostgresqlResult.lambda$map$0(PostgresqlResult.java:71)
	at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:100)
	at reactor.core.publisher.FluxZip$ZipCoordinator.drain(FluxZip.java:735)
	at reactor.core.publisher.FluxZip$ZipCoordinator.request(FluxZip.java:587)
```
The gist of my code is here:
https://gist.github.com/flyaruu/239a950ffda6f3fc544faffffc866f12

I have a toy database, with fake movies and fake actors, I query all movies, and for each movie I query all actors in that movie.
For up to about 20 movies that seems to work, after that it fails inside a codec.

The database is available on dockerhub:
https://hub.docker.com/r/dexels/dvdrental

It seems very similar to Issue #91

Squiry:
Actually database, qieru result size and codecs has nothing to do with that.
Simplified test case (for PostgresExample test class):
```java
@Test
void parallelTest() {
    Mono.from(getConnectionFactory().create())
        .flatMapMany(connection -> Flux.from(connection.createStatement("select 42").execute())
            .delayElements(Duration.ofMillis(1))
[...]

Issue #91, opened 2019-03-19
io.r2dbc.spi.Row.get(Object,Class<T>) fails when r2dbc-postgres is used with Kotlin coroutine.


mike-neck:
`io.r2dbc.spi.Row.get(Object,Class<T>)` fails when r2dbc-postgres is used with Kotlin coroutine.

Versions
---

* r2dbc-postgres : 1.0.0.M7
* Kotlin : 1.3.20
* kotlinx-coroutines-reactor: 1.1.1

Code
---

```kotlin
data class User(val id: Long, val name: String, val created: Instant)

@Test
fun r2dbcWithCoroutine() {
  val options = ConnectionFactoryOptions.builder()
      .option(ConnectionFactoryOptions.DRIVER, "postgresql")
      .option(ConnectionFactoryOptions.HOST, "localhost")
      .option(ConnectionFactoryOptions.PORT, 5432)
      .option(ConnectionFactoryOptions.USER, "postgres-user")
      .option(ConnectionFactoryOptions.PASSWORD, "postgres-pass")
      .option(ConnectionFactoryOptions.DATABASE, "postgres")
      .build()
  val connectionFactory = ConnectionFactories.get(options)
  val user = GlobalScope.mono {
    val conn = connectionFactory.create().awaitSingle()
    val result = conn.createStatement(
        //language=SQL
        "SELECT u.id, u.name, u.created FROM users AS u WHERE u.name = $1")
        .bind("$1", "test-user")
        .execute()
        .awaitSingle()
    return@mono result.map { row, _ -> // 62
      val name = row.get("name", String::class.java) // 63
      val id = row.get("id", Long::class.javaObjectType)
      val created = row.get("created", Instant::class.java)
      if (id != null && name != null && created != null) {
        return@map User(id, name, created)
      } else {
        return@map null
      }
    }.awaitSingle() // 71
  }
  runBlocking {
    println(user.awaitSingle()) // 75
  }
}
```

StackTrace
---

```
java.lang.StringIndexOutOfBoundsException
	at java.base/java.lang.String.checkBoundsOffCount(String.java:3304)
	at java.base/java.lang.String.<init>(String.java:505)
	at io.netty.buffer.ByteBufUtil.decodeString(ByteBufUtil.java:796)
	at io.netty.buffer.AbstractByteBuf.toString(AbstractByteBuf.java:1246)
[...]

Issue #86, opened 2019-03-13
Row.get seems to mix up columns


schauder:
I'm preparing an example for a conference talk.

I'm storing 1703 rows of data in a simple table
```
CREATE TABLE ROW( 
    ID SERIAL PRIMARY KEY, 
    CONTENT VARCHAR(2000)
)
```

When I try to load it again with: 
```
RowRepository rows;
\\....
    rows.findAll();
```
I get some rows back (typically ~30 sometimes up to ~150) before I get an:

```
java.lang.NumberFormatException: For input string: ""
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
```
(Full stack trace is below)

The fun part is that the string quoted seems most of the time two non-printable characters (boxes in the IDE console) but sometimes small bits of text like `"ink"` which leaves me with the impression that somehow data that belongs in one column ends up in another column.

```
org.springframework.data.mapping.MappingException: Couldn't read column id from Row.
	at org.springframework.data.r2dbc.function.convert.MappingR2dbcConverter$RowParameterValueProvider.getParameterValue(MappingR2dbcConverter.java:326) ~[spring-data-r2dbc-1.0.0.BUILD-20190310.103734-31.jar:1.0.0.BUILD-SNAPSHOT]
	at org.springframework.data.relational.core.conversion.BasicRelationalConverter$ConvertingParameterValueProvider.getParameterValue(BasicRelationalConverter.java:256) ~[spring-data-relational-1.1.0.BUILD-20190312.114323-483.jar:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.data.convert.ClassGeneratingEntityInstantiator$EntityInstantiatorAdapter.extractInvocationArguments(ClassGeneratingEntityInstantiator.java:250) ~[spring-data-commons-2.1.5.RELEASE.jar:2.1.5.RELEASE]
	at org.springframework.data.convert.ClassGeneratingEntityInstantiator$EntityInstantiatorAdapter.createInstance(ClassGeneratingEntityInstantiator.java:223) ~[spring-data-commons-2.1.5.RELEASE.jar:2.1.5.RELEASE]
	at org.springframework.data.convert.ClassGeneratingEntityInstantiator.createInstance(ClassGeneratingEntityInstantiator.java:84) ~[spring-data-commons-2.1.5.RELEASE.jar:2.1.5.RELEASE]
[...]

PR #116, opened 2019-05-24
Parallel mapping fix

Squiry:
CompositeByteBuf::discardReadComponents was called before mapping process in case of delayed mapping. 
Now we create new CompositeByteBuf from required parts for every envelope and that allows us to call discardReadComponents safe.

It should fix issues #107, #91 and #86.
#86 was tested with this [gist](https://gist.github.com/croudet/e34275f758765f7799b4870a45e8163d).
#107 was tested with this [gist](https://gist.github.com/flyaruu/239a950ffda6f3fc544faffffc866f12) and database from issue
#91 wasn't tested (need some help from someone with kotlin environment), but issue seems to be the same.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/message/backend/CopyData.java
+++ b/src/main/java/io/r2dbc/postgresql/message/backend/CopyData.java
@@ -77,7 +77,7 @@ public final class CopyData implements BackendMessage {
     static CopyData decode(ByteBuf in) {
         Assert.requireNonNull(in, "in must not be null");
 
-        return new CopyData(in.readSlice(in.readableBytes()));
+        return new CopyData(in.readSlice(in.readableBytes()).retain());
     }
 
 }
--- a/src/test/java/io/r2dbc/postgresql/message/backend/BackendMessageDecoderTest.java
+++ b/src/test/java/io/r2dbc/postgresql/message/backend/BackendMessageDecoderTest.java
@@ -17,71 +17,61 @@
 package io.r2dbc.postgresql.message.backend;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.CompositeByteBuf;
 import org.junit.jupiter.api.Test;
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Mono;
-import reactor.test.StepVerifier;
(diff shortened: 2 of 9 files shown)


src/main/java/io/r2dbc/postgresql/message/backend/CopyData.java
```diff
@@ -77,7 +77,7 @@ public String toString() {
     static CopyData decode(ByteBuf in) {
         Assert.requireNonNull(in, "in must not be null");
 
-        return new CopyData(in.readSlice(in.readableBytes()));
+        return new CopyData(in.readSlice(in.readableBytes()).retain());
```
Squiry: Actually it's a leak since CopyData has no release method, but I'm not sure if I should fix that in this pull request.

src/test/java/io/r2dbc/postgresql/message/backend/BackendMessageDecoderTest.java
```diff
@@ -17,71 +17,61 @@
...
     void authenticationGSS() {
-        decode('R', buffer -> buffer.writeInt(7))
-            .expectNext(AuthenticationGSS.INSTANCE)
-            .verifyComplete();
+        BackendMessage message = decode('R', buffer -> buffer.writeInt(7));
+        assertThat(message).isEqualTo(AuthenticationGSS.INSTANCE);
     }
 
     @Test
     void authenticationGSSContinue() {
```
Squiry: Every `BackendMessage` with `java.nio.ByteBuffer` fields can't be processed after `BackendMessageDecoder` called `release` on envelope buf since internal it's calling `ByteBuffer.duplicate` but clears the memory on calling `ByteBuf.release` (`UnpooledUnsafeDirectByteBuf.nioBuffer` and `UnpooledUnsafeDirectByteBuf.deallocate`).

msoub:
@mp911de @nebhale how could we move forward with this bug fix?

mp911de:
Thanks for the ping and sorry for the long radio silence. We want to merge this PR. It would be great to fix the memory issues. As you've spotted these, do you want to submit a PR for these as well? This time, it will not take _months_ to get these merged.

mp911de:
That's merged now.

PR #117, opened 2019-06-01
Support "postgres" driver

making:
As a Cloud Foundry user, I'd like to build a `ConnectionFactory` from [`DATABASE_URL`](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#DATABASE-URL) environment variable.

```java
ConnectionFactory connectionFactory = ConnectionFactories.get("r2dbc:" + System.getenv("DATABASE_URL"));
```

Currently r2dbc-postgresql supports only `postgresql` as a driver name but popular service brokers like elephantsql provides `postgres` schema like `postgres://exampleuser:examplepass@babar.elephantsql.com:5432/exampledb`

So, we need to replace `postgres` to `postgresql`

```java
ConnectionFactory connectionFactory = ConnectionFactories.get("r2dbc:" + System.getenv("DATABASE_URL").replace("postgres", "postgresql"));
```

It would be quite useful if r2dbc-postgresql accepts `postgres` and we can use `DATABASE_URL` directly.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
@@ -85,7 +85,7 @@ public final class PostgresqlConnectionFactoryProvider implements ConnectionFact
         Assert.requireNonNull(connectionFactoryOptions, "connectionFactoryOptions must not be null");
 
         String driver = connectionFactoryOptions.getValue(DRIVER);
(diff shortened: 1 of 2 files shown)


gregturn:
Resolved via https://github.com/r2dbc/r2dbc-postgresql/commit/47631f77326a6d0d33415b6bcaf89d7565a87725 and https://github.com/r2dbc/r2dbc-postgresql/commit/81a9f2f87f7c98ad17e45b3cfab21096c7a7ad2c.

Issue #118, opened 2019-06-02
Sql exceptions handling


vadim-shb:
I don't understand r2dbc exception handling policies
I have a snippet (r2dbc-postgresql[1.0.0.M7] + spring-data-r2dbc[1.0.0.M2]):

```
@Test
void test() {
    var connectionConfiguration = PostgresqlConnectionConfiguration.builder()
            .host("localhost")
            .database("db")
            .port(5432)
            .username("test")
            .password("test")
            .build();

    var connectionFactory = new PostgresqlConnectionFactory(connectionConfiguration);
    var databaseClient = DatabaseClient.create(connectionFactory);

    databaseClient.execute()
            .sql("CREATE TABLE foo (id VARCHAR(36) NOT NULL PRIMARY KEY)")
            .then().block();

    databaseClient.execute().sql("insert into foo(id) values ('2')").then().block();
    databaseClient.execute().sql("insert into foo(id) values ('2')").then().block();
}
```

I expect to get some kind of SQL-exception because of unique constraint violation, but instead, I just have
```
 ERROR | i.r.p.client.ReactorNettyClient | Error: SEVERITY_LOCALIZED=ERROR, SEVERITY_NON_LOCALIZED=ERROR, CODE=23505, MESSAGE=duplicate key value violates unique constraint "foo_pkey", DETAIL=Key (id)=(2) already exists., SCHEMA_NAME=public, TABLE_NAME=foo, CONSTRAINT_NAME=foo_pkey, FILE=nbtinsert.c, LINE=535, ROUTINE=_bt_check_unique
```
in logs and a hanging test...
Do I miss something or this is a bug?

Squiry:
You'll get your error if you try to subscribe on getRowsUpdated/map of Result. I don't know if it's 
expected behavior or not, the fix is pretty simple so we need a word from someone of r2dbc team.

schauder:
I tried to reproduce this but failed. I get the log message you mentioned plus an exception immediately after that. 

Could you create a testcase reproducing the issue? 

Also, at this point we should probably consider it a Spring Data R2DBC bug since that is the API you are using and there is no indication that this is specific to r2dbc-postgresql. 

`AbstractDatabaseClientIntegrationTests` is probably a good place to add a test.
But a standalone project on github would be just fine as well.

Squiry:
@schauder 
try this on master (insert it in `PostgresqlExample` class)
```java
    @Test
    void issue118() {
[...]

PR #119, opened 2019-06-05
No exception without calling getRowsUpdated/map on PostgresqlResult fix

Squiry:
Fix for #118


gregturn:
Resolved via https://github.com/r2dbc/r2dbc-postgresql/commit/2d24bcfca5fb5d8ce842720b08585612c2083bd1.

PR #140, opened 2019-08-29
Stub out the new autoCommit/isolationLevel methods.

gregturn:
Disable the related TCK test (for now).

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
@@ -132,6 +132,16 @@ public final class PostgresqlConnection implements Connection {
     }
 
     @Override
+    public IsolationLevel getTransactionIsolationLevel() {
+        return IsolationLevel.READ_COMMITTED; // TODO: Replace with proper lookup of transaction's isolation level.
+    }
+
+    @Override
+    public boolean isAutoCommit() {
+        return true; // TOODO: Replace with proper lookup of session's autoCommit status.
+    }
+
+    @Override
     public Mono<Void> releaseSavepoint(String name) {
         Assert.requireNonNull(name, "name must not be null");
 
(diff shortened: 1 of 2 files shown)


gregturn:
What's your opinion on doing this temporarily @nebhale @odrotbohm @davecramer @mp911de?

davecramer:
If this is blocking builds then +1

odrotbohm:
Isn’t the issue that it’s not even compiling due to the latest changes in the SPI?

gregturn:
Yes @odrotbohm.

odrotbohm:
Would you mind having a look at catching up with the SPI changes them. I guess we're better off building a proper implementation right away or at least use implementations that signal that the functionality is not available (signal errors instead of doing nothing).

I can't really judge how much time it'd take to build a proper implementation. So I'd love to get feedback on this one.

davecramer:
I'd agree that at the very least signalling errors would be a better way to go

killaitis:
My proposal is to add an implementation for setAutoCommit to fire up an SQL statement "SET AUTOCOMMIT = ON/OF" and ignore the isolation level issue for now. This PR not only blocks builds and further development, but the current SNAPSHOT build of r2dbc-postgres out in the wild is incompatible with the SNAPSHOT builds for -spi, pool etc. due to the missing members, causing other issues.

gregturn:
I already tested out writing that command. It doesn’t work. That command is part of ECPQL, not the actual server. 

It appears autocommit is a client thing not a server thing. You supposedly have to prefix each command with “BEGIN” to switch on autocommit behavior.

mp911de:
How does the JDBC driver work? It makes a lot of sense to stick to what JDBC does with Auto-commit behavior and isolation level retrieval.

davecramer:
Yes, that is more or less how we do it in JDBC. We remember if the session is in autocommit mode and if it is we do not issue a begin. If autocommit is false then a begin is prepended to the statement. https://github.com/pgjdbc/pgjdbc/blob/89464a54777b9272893ab24bc5e28a24623e41bb/pgjdbc/src/main/java/org/postgresql/jdbc/PgStatement.java#L410

gregturn:
I am tinkering with this on the branch and will continue next week. Learning the ropes of the project right now.

gregturn:
Superceded by #144.

Issue #139, opened 2019-08-28
PostgresqlConnection does not rollback if transaction status is FAILED


killaitis:
We recently ran into a problem where our connection pool was drained (all connections acquired, showing up in Postgres as "ROLLBACK" and "idle"). I was able to boil down the problem and it shows that all these stuck connections were caused by a data integrity exception thrown by r2dbc-postgres. 

Finally, I reduced the cause to a combination of 
- using a validation query for the pool
- having a corrupt connection after the data integrity exception ("current transaction is aborted, commands ignored until end of transaction block")

So, instead of being able to close the connection, close() will throw an exception, leaving the connection seemingly in a broken state.

The problem disappears when no validation query is used.

As far as I can analyse, this seems to be a regression since 0.8.0.M8.

Postgres: 11.4
r2dbc(-spi, -pool, -postgres): 0.8.0.BUILD-SNAPSHOT

Code to generate the issue:

```
    @Test
    void close_should_not_throw_exception() {
        ConnectionFactory connectionFactory = ConnectionFactories.get(ConnectionFactoryOptions.builder()
                .option(DRIVER, "postgresql")
                .option(HOST, "localhost")
                .option(USER, user)
                .option(PASSWORD, password)
                .option(DATABASE, database)
                .build());

        ConnectionPoolConfiguration configuration = ConnectionPoolConfiguration.builder(connectionFactory)
                .validationQuery("SELECT 1")
                .build();

        ConnectionPool connectionPool = new ConnectionPool(configuration);

        Connection connection = Mono.from(connectionPool.create()).block();

        Flux.from(connection.beginTransaction()).blockLast();
        execute(connection, "drop table if exists foo");
        execute(connection, "create table foo (id integer primary key)");
        execute(connection, "insert into foo (id) values (1)");

        assertThrows(PostgresqlServerErrorException.class, () -> {
            execute(connection, "insert into foo (id) values (1)");
        });

        Flux.from(connection.rollbackTransaction()).blockLast();

        assertDoesNotThrow(() -> {
            Flux.from(connection.close()).blockLast(); // This throws an unexpected exception
        });
    }

    static void execute(Connection connection, String sql) {
        Flux.from(connection.createStatement(sql).execute()).flatMap(Result::getRowsUpdated).blockLast();
    }
```

killaitis:
[...]

PR #141, opened 2019-09-01
Fixed rollback behavior in case of transaction status OPEN or FAILED

killaitis:
This is a fix for ticket #139. It will not compile without the patches from branch stubbed-out-spi, so builds and test will fail prior to merging it.


mp911de:
Thanks for report and your PR. We need to merge #140 before we can proceed with this one.

killaitis:
Rebased the patch to fix the failing tests. 
@mp911de any concerns merging the patch?

mp911de:
About to merge this one.

mp911de:
Thanks a lot, that's merged now.

PR #142, opened 2019-09-01
Add support for setting runtime parameters for new connections

killaitis:
This PR adds optional parameters that will be applied to each new DB connection by the create() method.

Use case: You have a project that is using a PostgreSQL database server, you need some server parameters to be modified for your client (e.g. the lock_timeout parameter) and you can't or don't want to change the server config to achieve this (hello, AWS Aurora!). You wand to apply these parameters to each new connection. You are using a connection pool and so you don't even easily know if a connection is new or pooled. Plus, you don't want to write such code at all, maybe your code is just using spring boot repositories. 

Idea: This is the exact same scenario as how the default schema is applied. It can be set in the connection configuration and it is applied to each new connection. 

Implementation: The solution is similar to the implementation of the schema parameter, but using a key/value map.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
@@ -21,6 +21,8 @@ import io.r2dbc.postgresql.util.Assert;
 import reactor.util.annotation.Nullable;
 
 import java.time.Duration;
+import java.util.Map;
+import java.util.Objects;
 
 /**
  * Connection configuration information for connecting to a PostgreSQL database.
@@ -42,6 +44,8 @@ public final class PostgresqlConnectionConfiguration {
 
     private final String host;
 
+    private final Map<String, String> options;
+
     private final String password;
 
     private final int port;
@@ -50,14 +54,15 @@ public final class PostgresqlConnectionConfiguration {
 
     private final String username;
 
-    private PostgresqlConnectionConfiguration(String applicationName, @Nullable Duration connectTimeout, @Nullable String database, boolean forceBinary, String host, String password, int port,
-                                              @Nullable String schema, String username) {
+    private PostgresqlConnectionConfiguration(String applicationName, @Nullable Duration connectTimeout, @Nullable String database, boolean forceBinary, String host,
+                                              @Nullable Map<String, String> options, String password, int port, @Nullable String schema, String username) {
 
         this.applicationName = Assert.requireNonNull(applicationName, "applicationName must not be null");
         this.connectTimeout = connectTimeout;
         this.database = database;
         this.forceBinary = forceBinary;
         this.host = Assert.requireNonNull(host, "host must not be null");
+        this.options = options;
         this.password = Assert.requireNonNull(password, "password must not be null");
         this.port = port;
         this.schema = schema;
@@ -81,6 +86,7 @@ public final class PostgresqlConnectionConfiguration {
             ", database='" + this.database + '\'' +
             ", forceBinary='" + this.forceBinary + '\'' +
             ", host='" + this.host + '\'' +
+            ", options='" + this.options + '\'' +
             ", password='" + this.password + '\'' +
             ", port=" + this.port +
             ", schema='" + this.schema + '\'' +
@@ -105,6 +111,9 @@ public final class PostgresqlConnectionConfiguration {
     String getHost() {
         return this.host;
     }
+
+    @Nullable
+    Map<String, String> getOptions() { return this.options; }
 
     String getPassword() {
         return this.password;
@@ -144,6 +153,8 @@ public final class PostgresqlConnectionConfiguration {
 
         private String host;
 
+        private Map<String, String> options;
+
         private String password;
 
         private int port = DEFAULT_PORT;
@@ -173,7 +184,7 @@ public final class PostgresqlConnectionConfiguration {
          * @return a configured {@link PostgresqlConnectionConfiguration}
          */
         public PostgresqlConnectionConfiguration build() {
-            return new PostgresqlConnectionConfiguration(this.applicationName, this.connectTimeout, this.database, this.forceBinary, this.host, this.password, this.port, this.schema, this.username);
+            return new PostgresqlConnectionConfiguration(this.applicationName, this.connectTimeout, this.database, this.forceBinary, this.host, this.options, this.password, this.port, this.schema, this.username);
         }
 
         public Builder connectTimeout(@Nullable Duration connectTimeout) {
@@ -212,6 +223,29 @@ public final class PostgresqlConnectionConfiguration {
          */
         public Builder host(String host) {
             this.host = Assert.requireNonNull(host, "host must not be null");
+            return this;
+        }
+
+        /**
+         * Configure connection initialization parameters.
+         *
+         * These parameters are applied once after creating a new connection. This is useful for setting up client-specific
+         * <a href="https://www.postgresql.org/docs/current/runtime-config-client.html#RUNTIME-CONFIG-CLIENT-FORMAT">runtime parameters</a>
+         * like statement timeouts, time zones etc.
+         *
+         * @param options the options
+         * @return this {@link Builder}
+         * @throws IllegalArgumentException if {@code options} or any key or value of {@code options} is {@code null}
+         */
+        public Builder options(Map<String, String> options) {
+            if (options == null) throw new IllegalArgumentException("options map must not be null");
+
+            options.forEach((k, v) -> {
+                Assert.requireNonNull(k, "option keys must not be null");
+                Assert.requireNonNull(v, "option values must not be null");
+            });
+
+            this.options = options;
             return this;
         }
 
@@ -257,6 +291,7 @@ public final class PostgresqlConnectionConfiguration {
                 ", database='" + this.database + '\'' +
                 ", forceBinary='" + this.forceBinary + '\'' +
                 ", host='" + this.host + '\'' +
+                ", parameters='" + this.options + '\'' +
                 ", password='" + this.password + '\'' +
                 ", port=" + this.port +
                 ", schema='" + this.schema + '\'' +
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactory.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactory.java
@@ -28,6 +28,8 @@ import io.r2dbc.postgresql.util.Assert;
 import io.r2dbc.spi.ConnectionFactory;
 import reactor.core.publisher.Mono;
 
+import java.util.Map;
+
 /**
  * An implementation of {@link ConnectionFactory} for creating connections to a PostgreSQL database.
  */
@@ -61,7 +63,7 @@ public final class PostgresqlConnectionFactory implements ConnectionFactory {
         return this.clientFactory
             .delayUntil(client ->
                 StartupMessageFlow
-                    .exchange(this.configuration.getApplicationName(), this::getAuthenticationHandler, client, this.configuration.getDatabase(), this.configuration.getUsername())
+                    .exchange(this.configuration.getApplicationName(), this::getAuthenticationHandler, client, this.configuration.getDatabase(), this.configuration.getUsername(), this.configuration.getOptions())
                     .handle(PostgresqlExceptionFactory::handleErrorResponse))
             .map(client -> new PostgresqlConnection(client, new DefaultCodecs(client.getByteBufAllocator()), DefaultPortalNameSupplier.INSTANCE, new IndefiniteStatementCache(client),
                 configuration.isForceBinary()))
@@ -71,6 +73,10 @@ public final class PostgresqlConnectionFactory implements ConnectionFactory {
     @Override
     public PostgresqlConnectionFactoryMetadata getMetadata() {
         return PostgresqlConnectionFactoryMetadata.INSTANCE;
+    }
+
+    PostgresqlConnectionConfiguration getConfiguration() {
+        return this.configuration;
     }
 
     @Override
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
@@ -16,12 +16,19 @@
 
 package io.r2dbc.postgresql;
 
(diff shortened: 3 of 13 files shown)


src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -215,6 +225,21 @@ public Builder host(String host) {
...
+         * Configure runtime options.
+         *
+         * These parameters are applied once after creating a new connection. This is useful for setting up client-specific
+         * <a href="https://www.postgresql.org/docs/current/runtime-config-client.html#RUNTIME-CONFIG-CLIENT-FORMAT">runtime parameters</a>
+         * like statement timeouts, time zones etc.
+         *
+         * @param options the options map
+         * @return this {@link Builder}
+         */
+        public Builder options(@Nullable Map<String, ?> options) {
```
mp911de: How about keeping `options` non-nullable? Reducing nullability helps to prevent potential NPE's in the first place.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactory.java
```diff
@@ -73,6 +76,10 @@ public PostgresqlConnectionFactoryMetadata getMetadata() {
         return PostgresqlConnectionFactoryMetadata.INSTANCE;
     }
 
+    public PostgresqlConnectionConfiguration getConfiguration() {
```
mp911de: Let's keep the configuration internal and not expose it.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -48,6 +54,16 @@
      */
     public static final Option<String> SCHEMA = Option.valueOf("schema");
 
+    /**
+     * Connection options which are applied once after the connection has been created.
+     * Format is '{key}={value},...,{key}={value}'
+     */
+    public static final Option<String> OPTIONS = Option.valueOf("options");
```
mp911de: Introducing `options` contradicts somewhat the purpose of the query-string/`ConnectionFactoryOptions ` itself. The idea of options is to keep config parameters on the top-level. I'd suggest to include a collection of known option names and then to iterate over these instead of extracting an options object.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -106,6 +111,9 @@ String getHost() {
         return this.host;
     }
 
+    @Nullable
+    Map<String, ?> getOptions() { return this.options; }
```
mp911de: Why `Map<String, ?>` and not `Map<String, String>`? Allowing objects as values introduce ambiguity over value accepted types. One could configure `options.put("statement_timeout", Duration.ofMinutes(1));` and expect this to work.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactory.java
killaitis: True. It was added for the tests to check the configuration, but I think making it package-private will be fine. Right?

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
killaitis: True. I'll change it.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
killaitis: That's true, but there are already tons of such options out there and each new release of PostgreSQL might introduce new options. Introducing constants for known option names introduces two problems IMO:

- it would require a new release of r2dbc-postgresql for new options introduced in PostgreSQL which adds a dependency one might not want. It requires you to update r2dbc-postgresql if you upgrade your DB server and want to use new options.
- the availability of these options as constants might let users think that they are generally available. Activating a (let's say) PG11 option in the ConnectionFactoryOptions but using a PG9 server will fail badly, possibly irritating the user. You would explicitly have to make it absolutely clear in the documentation which releases of PG are compatible. Combined with the hundreds of possible options already available… I don't know if this is a good idea.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
killaitis: That's true. The intention was to allow values of Integer, Long etc. but you're totally right. I'll change it to <String, String>

killaitis:
I've changed the issues discussed above.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
mp911de: PGJDBC has also an options object. Would it make sense to align R2DBC Postgres' `options` to what JDBC does? JDBC transports options in the startup packet which removes the need of running subsequent `SET` statements.

killaitis: I was able to modify the code to set the start options in the `StartupMessage` using the PGJDBC code as a template as proposed. Then I checked the official PostgreSQL documentation at https://www.postgresql.org/docs/current/protocol-message-formats.html which reads:

> options Command-line arguments for the backend. (This is deprecated in favor of setting individual run-time parameters.) 

So yes, I could move the connection initialisation parameters to the `StartupMessage`, but this is regarded as a deprecated way. What do you suggest? Go ahead moving it to `StartupMessage` or keep the implementation as-is?

mp911de: Thanks a lot. For database-specifics, I’m fan of doing what JDBC does. Maybe @davecramer can hint us in the right direction.

davecramer: I don't see anywhere where the jdbc driver uses options. It sends the parameters as parameters. https://github.com/pgjdbc/pgjdbc/blob/fcbbc3e6408cc1bcf459b740c683f3db40a5050c/pgjdbc/src/main/java/org/postgresql/core/v3/ConnectionFactoryImpl.java#L450

killaitis: And `options` is one of these parameters. It is pulled from the properties in lines 343-346 and merged into the parameters. There is a test case for this feature as well (https://github.com/pgjdbc/pgjdbc/blob/fcbbc3e6408cc1bcf459b740c683f3db40a5050c/pgjdbc/src/test/java/org/postgresql/test/core/OptionsPropertyTest.java). 

But as mentioned above, the `options` field is described as deprecated in the official PG docs.

davecramer: Fair enough, although deprecated does not mean it will be going away anytime soon. That said not using options is probably better

killaitis: Not using options means keeping the current implementation, i.e., sending a series of `SET` statements after creating the connection?

davecramer: Why send set for each option? The startup message can just contain all the parameters in one message.

killaitis: I think I do not fully get what you propose. You suggest setting the parameters in the startup message, but not to use options. But this (deprecated) options field in the startup message would be where to set them according to the protocol. This is also how it is implemented in pgjdbc.

davecramer: It is deprecated in favour of just sending parameters. You send name, value. JDBC encodes these into UTF-8 and each name, values is sent as a pair with a trailing 0. Options is not used. There is no need for multiple SET commands. See https://github.com/pgjdbc/pgjdbc/blob/fcbbc3e6408cc1bcf459b740c683f3db40a5050c/pgjdbc/src/main/java/org/postgresql/core/v3/ConnectionFactoryImpl.java#L476-L486

mp911de: So basically sending `options` as `-c search_path=foo` translates to a parameter key-value tuple with `search_path` as name and `foo` as value?

davecramer: Essentially, yes. And yes it is sent as a tuple.

killaitis: Ok, I think I get it now. :) I'll modify the code accordingly and push it for further review.

killaitis:
I changed the code as discussed. @davecramer @mp911de could you please have a look at the latest commit?

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -215,6 +225,22 @@ public Builder host(String host) {
...
+         * These parameters are applied once after creating a new connection. This is useful for setting up client-specific
+         * <a href="https://www.postgresql.org/docs/current/runtime-config-client.html#RUNTIME-CONFIG-CLIENT-FORMAT">runtime parameters</a>
+         * like statement timeouts, time zones etc.
+         *
+         * @param options the options
+         * @return this {@link Builder}
+         * @throws IllegalArgumentException if {@code options} is {@code null}
+         */
+        public Builder options(Map<String, String> options) {
+            this.options = Assert.requireNonNull(options, "options must not be null");
```
mp911de: Let's add here another assertion to make sure that `options` does not contain `null` keys or `null` values. Doing so prevents subsequent potential NPE's when sending the parameters.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -48,6 +55,16 @@
      */
     public static final Option<String> SCHEMA = Option.valueOf("schema");
 
+    /**
+     * Connection options which are applied once after the connection has been created.
+     * Format is '{key}={value};...;{key}={value}'
+     */
+    public static final Option<String> OPTIONS = Option.valueOf("options");
+
+    private static final Pattern itemPattern = Pattern.compile(";");
```
mp911de: I'm not exactly happy about nesting `options` and assuming a nested, encoded string as the resulting format would be:

```
r2dbc:postgres://…/database?options=lock_timeout%3D5s%3Bstatement_timeout%3D60000
```

After having a deeper look into the docs I totally understand why it's not feasible to include a pre-compiled property list.

How about not adding `options` to the URL and leaving options configuration for programmatic config only?

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
killaitis: Changed it.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -234,10 +235,13 @@ public Builder host(String host) {
          *
          * @param options the options
          * @return this {@link Builder}
-         * @throws IllegalArgumentException if {@code options} is {@code null}
+         * @throws IllegalArgumentException if {@code options} or any key or value of {@code options} is {@code null}
          */
         public Builder options(Map<String, String> options) {
-            this.options = Assert.requireNonNull(options, "options must not be null");
+            if (options == null) throw new IllegalArgumentException("options map must not be null");
+            if (options.keySet().stream().anyMatch(Objects::isNull)) throw new IllegalArgumentException("option keys must not be null");
```
mp911de: Note that `Stream` prohibits the use of `null` values. In general, we avoid using Java 8 streams because of these restrictions and performance issues. A `options.forEach((k, v) -> …)` seems more appropriate here.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
killaitis: Indeed, such a URL would be pretty nasty. 

But my primary vision was not to put the options into the URL, but rather use `R2dbcProperties.properties` and a separate entry the application.yml:

```
spring.r2dbc.properties.options: "lock_timeout=1s;statement_timeout=30min"
```

into my application.yml.

The ultimate goal would be to go one step further and remove the "options" nesting altogether:

```
spring.r2dbc.properties.lock_timeout: "1min"
spring.r2dbc.properties.statement_timeout: "30min"
```

This – then – could be also achieved using query parameters in the R2DBC URL.

But it would require some changes in the behaviour of `PostgresqlConnectionFactoryProvider`. Currently, the code only reads a select set of supported fields like `CONNECT_TIMEOUT` from the `ConnectionFactoryOptions`, so putting something into `spring.r2dbc.properties` or query parameters in `spring.r2dbc.url` is pointless ATM. Supporting them would require `PostgresqlConnectionFactoryProvider` to scan and use all options in  `ConnectionFactoryOptions` (and the class currently does not even allow option iteration as the options map is not exposed). 

If you like, I can move the options handling into this direction.

mp911de: Spring (Boot/Data) and the Postgres driver are two independent pieces. Let's try to address the configuration issue in a programmatic-way only by moving the problem into the config abstractions of Spring Boot by e.g. adding there some sort of `ConnectionFactoryOptionsCustomizer`.

In the driver, we would expect `options`  to be `Map<String, String>` (well, we cannot check for generics, but that's a different issue) instead of introducing a nested key-value format to the URL. The options constant would be defined as `public static final Option<Map<String, String>> OPTIONS = Option.valueOf("options");` and `PostgresqlConnectionFactoryProvider ` could reject non-`Map` values.

By doing so, we simplify the problem to some extent and allow augmentation of `ConnectionFactoryOptions` on a different level.

killaitis:
Rebased the patch to fix the failing tests. 
@mp911de any concerns merging the patch?

mp911de:
Thanks a lot. That's merged and polished now by cherry-picking to provide a proper commit message.

PR #161, opened 2019-09-17
Implement JSON and JSONB codecs

JoseLion:
Hi @mp911de! I'm opening this PR as an alternative to #39 but using [Jackson](https://github.com/FasterXML/jackson) dependency instead of GSON. I also took into account your suggestions to provide a certain level of flexibility and added return types for `ByteBuf`, `byte[]`, `String` and `InputStream`.

I only added the codec to `JSONB` since we'll need another codec for `JSON`. This is because the `doEncode(Object value)` method does not have any way to choose which one it should encode: JSON or JSONB.
I can open another PR after this one with the `JSON` codec or add it to this PR if you think it would be ok 🙂 

Cheers!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,8 @@ build/
 # virtual machine crash logs, see https://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 .flattened-pom.xml
+
+# Eclipse
+.classpath
+.project
+.settings
--- a/README.md
+++ b/README.md
@@ -187,7 +187,8 @@ This reference table shows the type mapping between [PostgreSQL][p] and Java dat
 | [`inet`][psql-inet-ref]                         | [**`InetAddress`**][java-inet-ref]|
 | [`integer`][psql-integer-ref]                   | [**`Integer`**][java-integer-ref], [`Boolean`][java-boolean-ref], [`Byte`][java-byte-ref], [`Short`][java-short-ref], [`Long`][java-long-ref], [`BigDecimal`][java-bigdecimal-ref]|
 | [`interval`][psql-interval-ref]                 | Not yet supported.|
-| [`json`][psql-json-ref]                         | Not yet supported.|
+| [`json`][psql-json-ref]                         | [`byte[]`][java-primitive-ref]|
+| [`jsonb`][psql-json-ref]                        | [`byte[]`][java-primitive-ref]|
 | [`line`][psql-line-ref]                         | Not yet supported.|
 | [`lseg`][psql-lseq-ref]                         | Not yet supported.|
 | [`macaddr`][psql-macaddr-ref]                   | Not yet supported.|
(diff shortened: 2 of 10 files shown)


mp911de:
Thanks for your PR. As stated in #39, supporting JSON(Oid) is something that we want. What we do not want is including a dependency to any JSON library in the core driver. We should expose JSON payload as binary (byte array, ByteBuffer, ByteBuf?, InputStream) data type, similar for writing. 

Any JSON processors should live on the client library or application level or can be contributed by driver extensions (driver extensibility is on our roadmap). 

We’re happy to merge basic support for now, without Jackson.

JoseLion:
@mp911de that makes sense! I updated the branch with the suggested approach, please let me know what you think 🙂 
I cannot change the branch name so if you prefer I could close this PR and open a new one with a proper branch name

mp911de:
No worries, the branch name isn’t an issue at all. Codecs already look well. Recently we kept binary and text formats within a single codec (see https://github.com/r2dbc/r2dbc-postgresql/blob/master/src/main/java/io/r2dbc/postgresql/codec/AbstractNumericCodec.java). I’d suggest adding an AbstractJsonCodec<T> with concrete subtypes for ByteBuffer, ByteBuf, byte[] and InputStream. 

The need for multiple subtypes is current a flaw of our approach that we will need to cleanup.

JoseLion:
@mp911de awesome, good to know we're going the right way! Just to be clear, we'll need the following codecs based on an `AbstractJsonCodec` right?
- `ByteBufferJsonCodec`, `ByteBufferJsonbCodec`
- `ByteBufJsonCodec`, `ByteBufJsonbCodec`
- `ByteArrayJsonCodec`, `ByteArrayJsonbCodec`
- `InputStreamJsonCodec`, `InputStreamJsonbCodec`

I'm not sure we can have `JSON` and `JSONB` in the same codec file since the `doEncode(Object value)` still doesn't have a way to know which one to use. Does this make sense? 🤔

mp911de:
Partially. `AbstractJsonCodec ` would resemble [`AbstractBinaryCodec`](https://github.com/r2dbc/r2dbc-postgresql/blob/master/src/main/java/io/r2dbc/postgresql/codec/AbstractBinaryCodec.java) to some extent. We don't require a distinction between `Json` and `Jsonb` codecs, they could be handled within the same codec. 

Codec decoding is based on format, type oid and the requested `Class`. Encoding solely relies on the value/value type. For `byte[]` and `ByteBuffer`, all encoding would be handled by our binary codecs as we don't have a mechanism to enforce a JSON codec to encode the value.

We want to introduce with R2DBC 0.9 type descriptors so a value can be encoded using a more specific codec.

mp911de:
Thanks for your PR. I'm going to take it from here.

mp911de:
There's one fundamental issue that became obvious during integration testing requiring us to rethink the API. Using `byte[]`, `ByteBuffer` or `String` as interchange type leads to:

```
 column "value" is of type jsonb but expression is of type bytea
```

as Postgres isn't able to convert between json/jsonb and bytea or varchar.

It seems that we should add a dedicated `Json` type so we can enforce a proper JSON type to encode outgoing values.

mp911de:
That's merged and polished now.

PR #164, opened 2019-09-20
Add CodecRegistrar extension point to register codecs

### LINKED ISSUE #95: Need a mechanism to dynamically register codecs for types after connection time
The driver currently handles decoding/encoding via an enum of well-known OIDs. For all the built-in types of Postgres, this is fine. However, types that come from extensions will not have a fixed OID value, they will be arbitrarily assigned by the database server when the extensions are installed. An common example of this is postgis.

The way the postgres jdbc driver handles this is that external libraries on the classpath can register decoders/encoders for types dynamically when the connection is created by name, and then when fetching the values as a generic Object, it can be cast to the specific type. These types are registered internally with a cache so that the OID lookups don't have to be done constantly, such as in here: https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jdbc/TypeInfoCache.java

I would propose that the `io.r2dbc.postgresql.PostgresqlConnectionFactory` should allow the end-user to supply their own Codecs factory. This way users can implement their own encoders/decoders for the custom types without breaking the contract of the API. I think the default implementation is fine and it should fit a majority of user's use-cases, but allowing it to be overridden would be extremely useful.

Also, I think that the `io.r2dbc.postgresql.codec.Codecs#preferredType` method should return a Publisher<Class<?>> so that `Codecs` implementations, if they so choose, can use the connection to query information about these extra types and do whatever work is necessary, such as querying the postgres information schemata to resolve types as needed.

mp911de:
The CodecRegistrar SPI can now discover extensions from the class path via Java's ServiceLoader mechanism using `META-INF/services/io.r2dbc.postgresql.extension.Extension`.

`CodecRegistrar` extension can contribute codecs for data types (OIDs) that are not supported by built-in driver codecs.

[resolves #95]


src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactory.java
```diff
@@ -101,6 +131,7 @@ PostgresqlConnectionConfiguration getConfiguration() {
     public String toString() {
         return "PostgresqlConnectionFactory{" +
             "clientFactory=" + this.clientFactory +
+            "codecRegistrars=" + this.extensions +
```
gregturn: Need a “,” here.

PR #165, opened 2019-09-22
Memory leak fixes

Squiry:
There's still one leak in TestKit.batch method: since nobody subscribes on "SELECT value FROM test" result PostgresqlRow.data just leaks, but I'm not sure is that a leak in a driver or wrong api usage.
Test can be written like this to prevent leak.
```java
@Test
default void batch() {
    getJdbcOperations().execute("INSERT INTO test VALUES (100)");

    Mono.from(getConnectionFactory().create())
        .flatMapMany(connection -> Flux.from(connection
            .createBatch()
            .add("INSERT INTO test VALUES(200)")
            .add("SELECT value FROM test")
            .execute())
            .concatWith(Mono.from(connection.close()).then(Mono.empty())))
        .as(StepVerifier::create)
        .consumeNextWith(r -> Mono.from(r.getRowsUpdated()).subscribe())
        .consumeNextWith(r -> Flux.from(r.map((row, meta) -> row.get(0))).subscribe())
        .verifyComplete();
}
```

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
@@ -73,10 +73,7 @@ public final class PostgresqlConnection implements Connection {
         this.statementCache = Assert.requireNonNull(statementCache, "statementCache must not be null");
         this.forceBinary = forceBinary;
         this.isolationLevel = Assert.requireNonNull(isolationLevel, "isolationLevel must not be null");
-        this.validationQuery = new SimpleQueryPostgresqlStatement(this.client, this.codecs, "SELECT 1")
-                .fetchSize(0)
-                .execute()
-                .flatMap(result -> result.map((row, meta) -> row.get(0, Integer.class)));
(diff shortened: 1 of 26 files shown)


src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
```diff
@@ -73,7 +73,10 @@
         this.statementCache = Assert.requireNonNull(statementCache, "statementCache must not be null");
         this.forceBinary = forceBinary;
         this.isolationLevel = Assert.requireNonNull(isolationLevel, "isolationLevel must not be null");
-        this.validationQuery = new SimpleQueryPostgresqlStatement(this.client, this.codecs, "SELECT 1").fetchSize(0).execute().flatMap(PostgresqlResult::getRowsUpdated);
+        this.validationQuery = new SimpleQueryPostgresqlStatement(this.client, this.codecs, "SELECT 1")
```
mp911de: Let's fix that issue directly in `PostgresqlResult.getRowsUpdated(…)`. Right now, `getRowsUpdated` filters for `CommandComplete` which does not release `DataRow`.

Also, we should not change formatting.

mp911de:

PR #179, opened 2019-10-12
Draft for logical decoding

mp911de:
We now support consumption of replication streams through logical and physical decoding. The API exposes functionality to create replication slots and to consume a replication stream.

```java
ReplicationSlotRequest request = ReplicationSlotRequest.logical()
    .slotName("my_slot")
    .outputPlugin("test_decoding")
    .temporary()
    .build();
Mono<ReplicationSlot> replicationSlot = replicationConnection.createSlot(request);

ReplicationRequest replicationRequest = ReplicationRequest.logical()
    .slotName(request.getSlotName())
    .startPosition(LogSequenceNumber.valueOf(0))
    .slotOption("skip-empty-xacts", true)
    .slotOption("include-xids", false)
    .build();
Mono<ReplicationStream> replicationStream = replicationConnection.startReplication(replicationRequest);

// Later

Flux<…> replicationMessages = replicationStream.map(byteBuf -> …);
Mono<Void> close = replicationStream.close();
```

TODO:

* [x] Request/value object naming
* [x] Documentation (readme)

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresReplicationStream.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresReplicationStream.java
@@ -1,0 +1,247 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.r2dbc.postgresql;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.util.ReferenceCountUtil;
+import io.r2dbc.postgresql.message.backend.BackendMessage;
+import io.r2dbc.postgresql.message.backend.CopyData;
+import io.r2dbc.postgresql.message.backend.ReadyForQuery;
+import io.r2dbc.postgresql.message.frontend.CopyDone;
+import io.r2dbc.postgresql.message.frontend.FrontendMessage;
+import io.r2dbc.postgresql.replication.LogSequenceNumber;
+import io.r2dbc.postgresql.replication.ReplicationRequest;
+import io.r2dbc.postgresql.replication.ReplicationStream;
+import io.r2dbc.postgresql.util.Assert;
+import io.r2dbc.spi.R2dbcNonTransientResourceException;
+import reactor.core.Disposable;
+import reactor.core.publisher.EmitterProcessor;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+
+final class PostgresReplicationStream implements ReplicationStream {
+
+    public static final long POSTGRES_EPOCH_2000_01_01 = 946684800000L;
+
+    private final EmitterProcessor<CopyData> responseProcessor = EmitterProcessor.create(false);
+
+    private final EmitterProcessor<FrontendMessage> requestProcessor;
+
+    private final AtomicReference<Disposable> subscription = new AtomicReference<>();
+
+    private final ByteBufAllocator allocator;
+
+    private final ReplicationRequest replicationRequest;
+
+    private final CompletableFuture<Void> closeFuture = new CompletableFuture<>();
+
+    private volatile LogSequenceNumber lastServerLSN = LogSequenceNumber.INVALID_LSN;
+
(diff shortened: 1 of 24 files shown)


src/main/java/io/r2dbc/postgresql/PostgresReplicationStream.java
```diff
@@ -0,0 +1,247 @@
...
+        ByteBuf byteBuf = prepareUpdateStatus(this.lastReceiveLSN, this.lastFlushedLSN, this.lastAppliedLSN, false);
+        io.r2dbc.postgresql.message.frontend.CopyData copyData = new io.r2dbc.postgresql.message.frontend.CopyData(byteBuf);
+        this.requestProcessor.onNext(copyData);
+    }
+
+    private ByteBuf prepareUpdateStatus(LogSequenceNumber received, LogSequenceNumber flushed,
+                                        LogSequenceNumber applied, boolean replyRequired) {
+        ByteBuf byteBuffer = this.allocator.buffer(34);
+
+        long now = System.currentTimeMillis();
```
davecramer: Don't use currentTimeMillis. NanoTime is more linear

davecramer:
@mp911de seems reasonable at first glance. Awesome!

src/main/java/io/r2dbc/postgresql/PostgresqlReplicationConnection.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.function.Predicate;
+
+import static io.r2dbc.postgresql.util.PredicateUtils.or;
+
+/**
+ * Postgres replication connection.
+ */
+final class PostgresqlReplicationConnection implements io.r2dbc.postgresql.api.PostgresqlReplicationConnection {
+
```
gregturn: Perhaps call it `DefaultPostgresqlReplicationConnection`, so you don't have to constant impl vs. API name collision and have to sprinkle your code with FQDN's.

src/main/java/io/r2dbc/postgresql/PostgresqlReplicationConnection.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public Mono<ReplicationSlot> createSlot(ReplicationSlotRequest request) {
+
+        Assert.requireNonNull(request, "request must not be null");
+
+        return this.delegate.createStatement(request.asSQL()).execute().flatMap(it -> {
+
+            return it.map((row, rowMetadata) -> {
+
+                return new ReplicationSlot(
```
gregturn: This too.

Frankly, seeing three nested returns statements is a bit scary. But maybe two nested lambdas is as well.

src/main/java/io/r2dbc/postgresql/PostgresqlReplicationConnection.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public Mono<ReplicationSlot> createSlot(ReplicationSlotRequest request) {
+
+        Assert.requireNonNull(request, "request must not be null");
+
+        return this.delegate.createStatement(request.asSQL()).execute().flatMap(it -> {
+
+            return it.map((row, rowMetadata) -> {
```
gregturn: You can replace this with a lambda. (I'd prefer it)

src/main/java/io/r2dbc/postgresql/message/frontend/CopyData.java
```diff
@@ -77,6 +78,16 @@ public boolean equals(Object o) {
...
     }
 
+    @Override
+    protected void deallocate() {
+        this.data.release();
+    }
+
+    @Override
+    public CopyData touch(Object hint) {
+        return this;
```
gregturn: Is this correct? A pass-through?

src/main/java/io/r2dbc/postgresql/replication/LogSequenceNumber.java
```diff
@@ -0,0 +1,118 @@
...
+    public static LogSequenceNumber valueOf(String strValue) {
+        int slashIndex = strValue.lastIndexOf('/');
+
+        if (slashIndex <= 0) {
+            return INVALID_LSN;
+        }
+
+        String logicalXLogStr = strValue.substring(0, slashIndex);
+        int logicalXlog = (int) Long.parseLong(logicalXLogStr, 16);
+        String segmentStr = strValue.substring(slashIndex + 1);
```
gregturn: I'd put a line break between `logicalXLogStr` and `segmentStr` creation.

src/main/java/io/r2dbc/postgresql/PostgresReplicationStream.java
```diff
@@ -0,0 +1,247 @@
...
+        if (disposable != null && this.subscription.compareAndSet(disposable, null)) {
+            disposable.dispose();
+
+            this.requestProcessor.onNext(CopyDone.INSTANCE);
+            this.requestProcessor.onComplete();
+
+            return this.responseProcessor.ignoreElements().then(Mono.fromCompletionStage(this.closeFuture));
+        }
+
+        return Mono.fromCompletionStage(this.closeFuture);
```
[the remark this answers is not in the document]
mp911de: The detour using `CompletableFuture` is to hand out a single handle instead that emits a value on completion. Using a `Publisher` requires additional complexity in the sense of having a close `Processor`. We can see multiple calls to `close()` and we want to notify also concurrent callers. At the same time, we want the close action to happen only once.

src/main/java/io/r2dbc/postgresql/message/frontend/CopyData.java
mp911de: `ReferenceCounted` API is made for fluent usage (`copyData.touch("foo").hashCode()`), see also https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java#L88-L97.

PR #199, opened 2019-11-25
Add OID to the NumericAbstractCodec SUPPORTED_TYPES

JoseLion:
This PR just adds the `OID` data type to the supported numeric types. This allows PostgreSQL's `OID` type to be decoded as `Integer`, `Long`, etc. According to PostgreSQL [documentation ](https://www.postgresql.org/docs/current/datatype-oid.html), the `OID` type is implemented as an unsigned  four-byte integer, so there should be no problem using the `NumericAbstractCodec` on this type 🙂


mp911de:
Care to add a codec test that shows that the OID type is properly decoded? A separate case in `AbstractCodecIntegrationTests` would do. Also, the `decodeNumber` should also handle the OID.

JoseLion:
Yep, sorry I forgot about tets 😅 and thanks for the advice on the `decodeNumber` method!

mp911de:

Issue #120, opened 2019-06-18
Postgresql failover cluster missing functionality


mangusbrother:
Currently this is unable to support a master-slave scenario for postgresql

basically the equivalent of this:

`jdbc:postgresql://postgres_master:5432,postgres_slave:5432/myDb`

mp911de:
Happy to review a pull request on how to make `Client` customizable for various clustering- and failover strategies.

cambierr:
Here is something quick and dirty I did, given two `ConnectionFactory` objects bound to respectively slave and master servers, I implemented a custom `ConnectionFactory` wrapping the two others, and the most interesting part is:

```java
    @Override
    public Publisher<? extends Connection> create() {
        return TransactionSynchronizationManager
                .forCurrentTransaction()
                .flatMap(a -> a.isCurrentTransactionReadOnly() ? slavePool.create() : masterPool.create())
                .onErrorResume(NoTransactionException.class, e -> slavePool.create());
    }
```

Do you guys have an opinion on this (again, this is quick and dirty) ? If you think this is going in the good direction, we may open a PR with a highly-polished version of this custom `ConnectionFactory`

Elvirka:
Hello, we have a problem with setting postgresql targetServerType. When we use it in our connection string, we get an exceptionio.r2dbc.postgresql.ExceptionFactory$PostgresqlBadGrammarException: unrecognized configuration parameter "targetServerType"

davecramer:
> Hello, we have a problem with postgresql setting targetServerType. When we use it in our connection string, we get an exception
> io.r2dbc.postgresql.ExceptionFactory$PostgresqlBadGrammarException: unrecognized configuration parameter "targetServerType"

This does not exist as of yet. This is a JDBC thing

kotoant:
Hi, I've just written FailoverConnectionFactoryProvider.
It works fine for me.

```
package kotoant;

import io.r2dbc.spi.Connection;
import io.r2dbc.spi.ConnectionFactories;
import io.r2dbc.spi.ConnectionFactory;
import io.r2dbc.spi.ConnectionFactoryMetadata;
import io.r2dbc.spi.ConnectionFactoryOptions;
import io.r2dbc.spi.ConnectionFactoryProvider;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import reactor.core.publisher.Flux;
import reactor.core.publisher.Mono;

import java.util.ArrayList;
import java.util.Collections;
[...]

PR #203, opened 2019-11-27
Failover functionality

Squiry:
Implementation of functionality described in [this](https://github.com/r2dbc/r2dbc-postgresql/issues/120) issue.


src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -594,8 +537,129 @@ public Builder username(String username) {
...
+            return this;
+        }
+
+        /**
+         * Controls how long in seconds the knowledge about a host state is cached connection factory. The default value is 10000 milliseconds.
+         *
+         * @param hostRecheckTime host recheck time in milliseconds
+         * @return this {@link Builder}
+         */
+        public Builder hostRecheckTime(int hostRecheckTime) {
```
mp911de: Nit: Let's use `Duration` here

src/main/java/io/r2dbc/postgresql/TargetServerType.java
```diff
@@ -0,0 +1,53 @@
+package io.r2dbc.postgresql;
+
+import javax.annotation.Nullable;
+
+public enum TargetServerType {
```
mp911de: Should we convert `TargetServerType` to an `abstract class`/`interface` + default `enum` so that users can implement their own logic into `allowStatus`?

src/main/java/io/r2dbc/postgresql/TargetServerType.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final String value;
+
+    TargetServerType(String value) {
+        this.value = value;
+    }
+
+    @Nullable
+    public static TargetServerType fromValue(String value) {
+        String fixedValue = value.replace("lave", "econdary");
```
mp911de: Nit: Typo

mp911de:
@Squiry can you push your code into a branch in this repo so we can collaborate on it together?

ankur4u007:
@mp911de is there any update on the implementing this failover(master and slave) functionality, I see this PR is stalled in favor of new ground up implementation per #250. I wanted to know if someone is actively working on it and its planned for any release or it's still open for contribution ?
Thanks!

mp911de:
So far there was no movement. Given the numerous flavors of servers that use the Postgres protocol (Postgres with replication, Yugabyte, Cockroach, ...) we aim for a simple and pluggable client/connect model. Feel free to pick up on that end. /cc @Squiry

jgrgt:
@mp911de Any updates on this? We would like to use this kind of functionality to ensure quick recovery/failover. I understand a preference was given for a 'simpler and pluggable client/connect model'. Could you explain more what that would mean? What would be required?

We would really like the functionality offered by the old PR's to allow us to spread load over a cluster of Aurora read nodes and allow us to quickly recover from read node failures...

kotoant:
Here is the possible solution of the problem:
https://github.com/pgjdbc/r2dbc-postgresql/issues/120#issuecomment-765471935

mp911de:
Thanks, @Squiry for your initial work. We picked it up in a follow-up pull request. Superseded by #474

PR #205, opened 2019-12-01
Fix typo in project url

DeKaN:
Fix project URL typos in pom.xml

@pivotal-issuemaster This is an Obvious Fix


mp911de:

PR #210, opened 2019-12-09
Simplify factory provider

ask4gilles:
proposal to simplify the ```PostgresqlConnectionFactoryProvider```

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
@@ -16,7 +16,6 @@
 
 package io.r2dbc.postgresql;
 
-
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.r2dbc.postgresql.client.DefaultHostnameVerifier;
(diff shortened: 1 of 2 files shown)


mp911de:
Thanks for your pull request. Moving the initialization into the `Builder` is an interesting option that would allow applications to pre-initialize the builder with `ConnectionFactoryOptions`. We probably should inline parts of the methods (such as not having an individual method per SSL option).

It would also make sense to expose `builder(ConnectionFactoryOptions)` statically to obtain an initialized builder instead of a method that applies `ConnectionFactoryOptions` to an existing builder. Applying some other configuration object bears always the risk that previously configured options get overwritten and by returning a pre-initialized builder, we're able to avoid this risk.

ask4gilles:
Hi Mark,
I've pushed some modifications according to your remarks.

mp911de:
Thanks a lot. While reviewing the code I realized that moving the initialization `ConnectionFactoryOptions`-based initialization to `PostgresqlConnectionConfiguration.Builder` causes a cycle between `PostgresqlConnectionConfiguration.Builder` and `PostgresqlConnectionFactoryProvider` as driver-specific options are declared in `PostgresqlConnectionFactoryProvider`.

I'm going to move the entire `fromConnectionFactoryOptions(…)` method tree back to `PostgresqlConnectionFactoryProvider` preserving the optimizations from the original pull request.

mp911de:
That's merged and polished now.

ask4gilles:
Thanks Mark!

Issue #212, opened 2019-12-16
On error not triggered when listen connection terminated


steven-sheehy:
## Bug Report

Fails to trigger an `onError` when the database is stopped and using postgresql listen. 

#### Versions

- Driver: `0.8.0.RELEASE`
- Database: `postgres:11.3-alpine`
- Java: 11
- OS: MacOS

#### Current Behavior

<details>
<summary>Stack trace</summary>

```
// your stack trace here
```

</details>
 
#### Table schema

<details>
<summary>Input Code</summary>

```sql
-- your SQL here;
```

</details>

#### Steps to reproduce

Run JUnit test: `R2dbcListenApplicationTests.testDisconnect()`:
https://github.com/steven-sheehy/r2dbc-listen

<details>
<summary>Input Code</summary>

```java
// your code here;
```

</details>

#### Possible Solution

- Trigger an onError immediately
- Attempt reconnect automatically within some timeout, then trigger an onError if unsuccessful

mp911de:
Probably it would make sense to send different signals when the connection was closed intentionally and unintentionally. If the server goes down, we should send `onError`. When the application gets shut down, then we would complete the stream normally (`onComplete`).

mp911de:
That's fixed now.

PR #227, opened 2020-01-20
Terminate Notification stream on disconnect

mp911de:
We now terminate the notification stream (PostgresqlConnection.getNotifications()) when the connection gets disconnected. If the connection is closed normally, the stream terminates successfully (onComplete). Unintended disconnects result in an error (onError).

Requires forward-port to `master` and `0.8.x.`.

[#212]


gregturn:
@mp911de I have tried rebasing this against latest `0.8.x` but there are too many changes to `ReactorNettyClient` that I don't 100% grok.

gregturn:
I rebased the PR. After cleaning up `ReactorNettyClient`, it still doesn't compile due to a new static class now pointed at a non-static field. It also has two versions of `exchange()`, neither of which I can discern is the "correct" one. (Perhaps the correct version is a combination of the two?)

If you can fix it, then perhaps rebasing/squashing is in order.

PR #225, opened 2020-01-08
Prepared statement cache configuration

Squiry:
https://github.com/r2dbc/r2dbc-postgresql/issues/223

Added preparedStatementCacheQueries parameter, which controls cache behaviour.
When cache limit is reached last used statement is closed before parsing new one.
When cache limit is 0 caching of statement is disabled and unnamed statements are used in extended message flow.
Negative values stand for IndefiniteStatementCache (default one).

PR #228, opened 2020-01-21
Add DiscardOnCancel operator

### LINKED ISSUE #222: Insertion of too many rows do not complete
#### Versions

- Driver: 0.8.0.RELEASE
- Database: postgres:9.0/postgres:12.1/postgres:12.1-alpine
- Java: 1.8/11/13
- OS: Windows/MacOS

#### Current Behavior

Insertion queries that use bind and returning do not complete when inserting too many rows.

When inserting multiple rows, there is a limit to the amount that can be inserted. When that limit is reached, the execution will just hang forever.

The problem is happening with and without the use of the transactional operator. The only difference between them is the amount of rows inserted before the issue happens.
When not using transactional operator, 518 rows and then it will hang.
When using transactional operator, 260 rows will be inserted and then it will hang.

Those numbers do also change dependending on the queries that were run before the issue query. This behavior can be seen by changing the last argument passed to the method called "behaviorChangingSample" that is present in the repro code. In general, the amount of inserts with the issue query before the issue will be increased by 2 by each query ran before it.

There is no error stack trace at all, it just hangs.

</details>
 
#### Table schema

<details>
<summary>Input Code</summary>

```sql
CREATE TABLE foo
(
    id    SERIAL PRIMARY KEY,
    value CHAR(1) NOT NULL
);
```

</details>

#### Steps to reproduce

Just execute the code below or the DemoApplication in the repo below.
There are 5 methods in the repro code. They do showcase that the problem is only hapenning in insert queries using bind and return at the same time.

Repo link: https://github.com/gabrieldn/spring-r2dbc-postgres-insert-issue

<details>
<summary>Input Code</summary>

```java
import io.r2dbc.pool.ConnectionPool;
import io.r2dbc.pool.ConnectionPoolConfiguration;
import io.r2dbc.spi.ConnectionFactories;
import io.r2dbc.spi.ConnectionFactory;
import io.r2dbc.spi.ConnectionFactoryOptions;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.data.r2dbc.connectionfactory.R2dbcTransactionManager;
import org.springframework.data.r2dbc.core.DatabaseClient;
import org.springframework.transaction.ReactiveTransactionManager;
import org.springframework.transaction.reactive.TransactionalOperator;
import org.testcontainers.containers.PostgreSQLContainer;

import java.time.Duration;
import java.util.ArrayList;
import java.util.List;

public class DemoApplication {

    private static final Logger LOGGER = LoggerFactory.getLogger(DemoApplication.class);

    public static void main(String[] args) {
        ConnectionFactory connectionFactory = createConnectionFactory();
        DatabaseClient client = createDatabaseClient(connectionFactory);
        ReactiveTransactionManager tm = createTransactionManager(connectionFactory);

        client.execute("create table foo (id serial primary key, value char(1) not null)")
                .fetch().rowsUpdated().block();

        TransactionalOperator operator = TransactionalOperator.create(tm);

//        LOGGER.info("rawInsertString inserted {} rows", rawInsertString(client, operator, 5000).size());
//        LOGGER.info("insertWithBind inserted {} rows", insertWithBind(client, operator, 5000).size());
//        LOGGER.info("rawInsertStringWithReturning inserted {} rows", rawInsertStringWithReturning(client, operator, 5000).size());
        LOGGER.info("insertWithBindAndReturning inserted {} rows", insertWithBindAndReturning(client, operator).size());

//        behaviorChangingSample(client, operator, 1);
    }

    private static List<String> insertWithBind(DatabaseClient client, TransactionalOperator operator, long rowsToInsert) {
        return client.execute("INSERT INTO foo(value) VALUES (:value)")
                .bind("value", "x")
                .fetch().first()
                .repeat(rowsToInsert - 1).then()
                .as(operator::transactional)
                .then(client.execute("select * from foo")
                        .map((row, rowMetadata) -> String.valueOf(row.get("id")))
                        .all().collectList())
                .defaultIfEmpty(new ArrayList<>())
                .block();
    }

    private static List<String> rawInsertString(DatabaseClient client, TransactionalOperator operator, long rowsToInsert) {
        return client.execute("INSERT INTO foo(value) VALUES ('x')")
                .fetch().first()
                .repeat(rowsToInsert - 1).then()
                .as(operator::transactional)
                .then(client.execute("select * from foo")
                        .map((row, rowMetadata) -> String.valueOf(row.get("id")))
                        .all().collectList())
                .defaultIfEmpty(new ArrayList<>())
                .block();
    }

    private static List<String> rawInsertStringWithReturning(DatabaseClient client, TransactionalOperator operator, long rowsToInsert) {
        return client.execute("INSERT INTO foo(value) VALUES ('x') RETURNING *")
                .map((row, metadata) -> {
                    LOGGER.info("rawInsertStringWithReturning Id: {}", row.get("id"));
                    return row;
                }).first()
                .repeat(rowsToInsert - 1).then()
                .as(operator::transactional)
                .then(client.execute("select * from foo")
                        .map((row, rowMetadata) -> String.valueOf(row.get("id")))
                        .all().collectList())
                .defaultIfEmpty(new ArrayList<>())
                .block();
    }

    private static List<String> insertWithBindAndReturning(DatabaseClient client, TransactionalOperator operator) {
        return client.execute("INSERT INTO foo(value) VALUES (:value) RETURNING *")
                .bind("value", "x")
                .map((row, metadata) -> {
                    LOGGER.info("insertWithBindAndReturning: {}", row.get("id"));
                    return row;
                }).first()
                .repeat(5000 - 1).then()
                .as(operator::transactional)
                .then(client.execute("select * from foo")
                        .map((row, rowMetadata) -> String.valueOf(row.get("id")))
                        .all().collectList())
                .defaultIfEmpty(new ArrayList<>())
                .block();
    }

    private static void behaviorChangingSample(DatabaseClient client, TransactionalOperator operator, long rowsToInsert) {
        LOGGER.info("rawInsertString inserted {} rows", insertWithBind(client, operator, rowsToInsert).size());
        LOGGER.info("insertWithBindAndReturning inserted {} rows", insertWithBindAndReturning(client, operator).size());
    }

    private static PostgreSQLContainer<?> createContainer() {
        PostgreSQLContainer<?> container = new PostgreSQLContainer<>("postgres:12.1-alpine");
        container.start();

        return container;
    }

    private static ConnectionFactory createConnectionFactory() {
        PostgreSQLContainer<?> container = createContainer();

        ConnectionFactory connectionFactory = ConnectionFactories.get(ConnectionFactoryOptions.builder()
                .option(ConnectionFactoryOptions.DRIVER, "pool")
                .option(ConnectionFactoryOptions.PROTOCOL, "postgres")
                .option(ConnectionFactoryOptions.HOST, container.getContainerIpAddress())
                .option(ConnectionFactoryOptions.PORT, container.getMappedPort(5432))
                .option(ConnectionFactoryOptions.USER, container.getUsername())
                .option(ConnectionFactoryOptions.PASSWORD, container.getPassword())
                .option(ConnectionFactoryOptions.DATABASE, container.getDatabaseName())
                .build());

        ConnectionPoolConfiguration configuration = ConnectionPoolConfiguration.builder(connectionFactory)
                .maxIdleTime(Duration.ofMinutes(10))
                .maxLifeTime(Duration.ofMinutes(30))
                .initialSize(2)
                .maxSize(5)
                .build();

        return new ConnectionPool(configuration);
    }

    private static DatabaseClient createDatabaseClient(ConnectionFactory connectionFactory) {
        return DatabaseClient.create(connectionFactory);
    }

    private static ReactiveTransactionManager createTransactionManager(ConnectionFactory connectionFactory) {
        return new R2dbcTransactionManager(connectionFactory);
    }
}
```

</details>

#### Expected behavior/code

Insert as many rows as needed.

#### Additional context

The problem seems to be happening only with the postgres driver. I have tested with both mysql drivers available and the h2 driver and it works just fine even when inserting a few million rows.

When running with spring webflux, the problem does get worse and worse. Sometimes being unable to insert just a few rows, the worst case I got was failing to complete a POST request that should insert 10 rows.

mp911de:
FluxDiscardOnCancel replays source signals unless cancelling
the subscription. On cancellation, the subscriber requests Long.MAX_VALUE
to drain the source and discard elements that are emitted afterwards.

[closes #222]

/cc @squiry

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/ExtendedQueryPostgresqlStatement.java
+++ b/src/main/java/io/r2dbc/postgresql/ExtendedQueryPostgresqlStatement.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 
 package io.r2dbc.postgresql;
 
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
 import io.r2dbc.postgresql.api.PostgresqlStatement;
 import io.r2dbc.postgresql.client.Binding;
 import io.r2dbc.postgresql.client.Client;
@@ -28,6 +30,7 @@ import io.r2dbc.postgresql.message.backend.CloseComplete;
 import io.r2dbc.postgresql.message.backend.NoData;
 import io.r2dbc.postgresql.util.Assert;
 import io.r2dbc.postgresql.util.GeneratedValuesUtils;
+import io.r2dbc.postgresql.util.Operators;
 import reactor.core.publisher.Flux;
 
 import java.util.ArrayList;
@@ -182,11 +185,16 @@ final class ExtendedQueryPostgresqlStatement implements PostgresqlStatement {
 
         ExceptionFactory factory = ExceptionFactory.withSql(sql);
         return this.statementCache.getName(this.bindings.first(), sql)
-            .flatMapMany(name -> ExtendedQueryMessageFlow
-                .execute(Flux.fromIterable(this.bindings.bindings), this.client, this.portalNameSupplier, name, sql, this.forceBinary))
+            .flatMapMany(name -> {
+                return ExtendedQueryMessageFlow
+                    .execute(Flux.fromIterable(this.bindings.bindings), this.client, this.portalNameSupplier, name, sql, this.forceBinary);
+            })
             .filter(RESULT_FRAME_FILTER)
             .windowUntil(CloseComplete.class::isInstance)
-            .map(messages -> PostgresqlResult.toResult(this.codecs, messages, factory));
+            .map(messages -> PostgresqlResult.toResult(this.codecs, messages, factory))
+            .cast(io.r2dbc.postgresql.api.PostgresqlResult.class)
+            .as(Operators::discardOnCancel)
+            .doOnDiscard(ReferenceCounted.class, ReferenceCountUtil::release);
     }
 
     private int getIndex(String identifier) {
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2019 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import io.r2dbc.postgresql.client.SimpleQueryMessageFlow;
 import io.r2dbc.postgresql.client.TransactionStatus;
 import io.r2dbc.postgresql.codec.Codecs;
 import io.r2dbc.postgresql.util.Assert;
+import io.r2dbc.postgresql.util.Operators;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.IsolationLevel;
 import io.r2dbc.spi.ValidationDepth;
@@ -334,6 +335,7 @@ final class PostgresqlConnection implements io.r2dbc.postgresql.api.PostgresqlCo
 
     private Mono<Void> useTransactionStatus(Function<TransactionStatus, Publisher<?>> f) {
         return Flux.defer(() -> f.apply(this.client.getTransactionStatus()))
+            .as(Operators::discardOnCancel)
             .then();
     }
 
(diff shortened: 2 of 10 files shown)


Squiry:
I believe my changes [here](https://github.com/r2dbc/r2dbc-postgresql/blob/master/src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java#L142) kind of almost fixed this issue.
Your tests are not failing on master, but there's an obvious leak that can be fixed without all those operators, we just need to change that:
```java
receiver.sink.next(message);
```
to that
```java
if (receiver.sink.isCancelled()) {
    ReferenceCountUtil.release(message);
} else {
    receiver.sink.next(message);
}
```

mp911de:
It makes sense to fix the loophole. There is another issue here that we should address: `FluxDiscardOnCancel` works on a proper backpressure implementation. The blocked statement issue was caused because the demand (as seen by Netty Inbound) went to zero. 

Now with `handle((message, sink) -> {` (introduced with #225), we now no longer bind the backpressure from the R2DBC SPI to the netty channel which can overflow the actual `sink`.

PR #235, opened 2020-01-29
Add codec for BigInteger

### LINKED ISSUE #233: Add a codec for BigInteger
I would like to add support for the `BigInteger` type to the codecs.

In our project we use r2dbc in combination with code generated by jooq (based on the sql schema). When a Numeric column has a scale of 0 (the default) and has a sufficiently high precision, the generated java type for the column is a `BigInteger`. 

I would like to be able to get the column value as this type from r2dbc by adding a codec similar to the `BigDecimalCodec`.

I have currently written this codec in our own project but would like to add them to r2dbc-postgresql.

timve:
Applicable to any numeric type with scale 0

[resolves #233]

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/BigIntegerCodec.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/BigIntegerCodec.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


src/main/java/io/r2dbc/postgresql/codec/BigIntegerCodec.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public Parameter encodeNull() {
+        return createNull(NUMERIC, FORMAT_TEXT);
+    }
+
+    @Override
+    BigInteger doDecode(ByteBuf buffer, PostgresqlObjectId dataType, @Nullable Format format, @Nullable Class<? extends BigInteger> type) {
+        Assert.requireNonNull(buffer, "byteBuf must not be null");
+
+        return this.decodeNumber(buffer, dataType, format, BigInteger.class, it -> new BigInteger(it.toString()));
```
timve: This code potentially throws `NumberFormatException` when the scale is > 0. 
I did not find any defensive code in the other subclasses of `AbstractNumericCodec` so I did not add checks to see if construction of the `BigInteger` is possible using the decoded value.

mp911de: Our codecs downcast values if they are requested to do so. E.g. a `FLOAT8` column result can be consumed as `Double`, `Long`, `Int` etc. I think we should do the same here and return the `BigInteger` value.

I will take care of this aspect during the merge.

PR #245, opened 2020-02-08
Fetch size implementation

Squiry:
Fetch size implementation for ExtendedQueryPostgresqlStatement.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/client/ExtendedQueryMessageFlow.java
+++ b/src/main/java/io/r2dbc/postgresql/client/ExtendedQueryMessageFlow.java
@@ -17,11 +17,15 @@
 package io.r2dbc.postgresql.client;
 
 import io.netty.buffer.Unpooled;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
 import io.r2dbc.postgresql.message.Format;
 import io.r2dbc.postgresql.message.backend.BackendMessage;
 import io.r2dbc.postgresql.message.backend.CloseComplete;
+import io.r2dbc.postgresql.message.backend.CommandComplete;
 import io.r2dbc.postgresql.message.backend.ErrorResponse;
 import io.r2dbc.postgresql.message.backend.ParseComplete;
+import io.r2dbc.postgresql.message.backend.PortalSuspended;
 import io.r2dbc.postgresql.message.backend.ReadyForQuery;
 import io.r2dbc.postgresql.message.frontend.Bind;
 import io.r2dbc.postgresql.message.frontend.Close;
@@ -33,12 +37,15 @@ import io.r2dbc.postgresql.message.frontend.FrontendMessage;
 import io.r2dbc.postgresql.message.frontend.Parse;
 import io.r2dbc.postgresql.message.frontend.Sync;
 import io.r2dbc.postgresql.util.Assert;
-import org.reactivestreams.Publisher;
+import io.r2dbc.postgresql.util.Operators;
+import reactor.core.publisher.DirectProcessor;
 import reactor.core.publisher.Flux;
-import reactor.core.publisher.Mono;
+import reactor.core.publisher.FluxSink;
+import reactor.core.publisher.SynchronousSink;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
 
 import static io.r2dbc.postgresql.message.frontend.Execute.NO_LIMIT;
@@ -60,7 +67,7 @@ public final class ExtendedQueryMessageFlow {
     /**
      * Execute the execute portion of the <a href="https://www.postgresql.org/docs/current/static/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY">Extended query</a> message flow.
      *
-     * @param bindings           the {@link Binding}s to bind
+     * @param binding            the {@link Binding} to bind
      * @param client             the {@link Client} to exchange messages with
      * @param portalNameSupplier supplier unique portal names for each binding
      * @param statementName      the name of the statementName to execute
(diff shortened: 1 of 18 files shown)


src/main/java/io/r2dbc/postgresql/util/Assert.java
```diff
@@ -133,4 +134,28 @@ public static String requireFileExistsOrNull(@Nullable String file, String messa
...
+     *
+     * @param value     the value to check
+     * @param predicate predicate to check value
+     * @param message   the message to use in exception if type is not as required
+     * @param <T>       the type being required
+     * @return the value casted to the required type
+     * @throws IllegalArgumentException if {@code value} is not of the required type
+     * @throws IllegalArgumentException if {@code value}, {@code type}, or {@code message} is {@code null}
+     */
+    public static <T> T require(T value, Predicate<T> predicate, String message) {
```
mp911de: Typically, we have something like [`Assert.isTrue(…)`](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/Assert.java#L116). Spring 5 introduced also variants that accept a `Supplier<String> messageSupplier` so that messages using `String.format(…)` don't get executed each time the method gets hit.

src/main/java/io/r2dbc/postgresql/client/ExtendedQueryMessageFlow.java
```diff
@@ -69,15 +76,55 @@ private ExtendedQueryMessageFlow() {
...
+                    .concatWithValues(Sync.INSTANCE);
+                return client.exchange(bindFlow)
+                    .as(Operators::discardOnCancel)
+                    .doOnDiscard(ReferenceCounted.class, ReferenceCountUtil::release);
+            });
+        }
+        return Flux.defer(() -> {
+            String portal = portalNameSupplier.get();
+            Flux<FrontendMessage> bindFlow = toBindFlow(binding, portal, statementName, query, forceBinary, fetchSize).concatWithValues(Flush.INSTANCE);
+            DirectProcessor<FrontendMessage> requestsProcessor = DirectProcessor.create();
```
mp911de: How about using just `EmitterProcessor` without an additional sink? We can avoid another queue creation with `EmitterProcessor`.

Squiry: `EmitterProcessor` has some queue inside, `DirectProcessor` has no. We should never have situation when `DirectProcessor` has no subscribers.

Squiry:
Simple queries have no Execute(portal, n) part, so we cannot set fetch size on them. We can create another method on `PostgresqlConnection`, something like `createExtendedStatement` and tweak ExtendedStatement flow to work with empty bindings, protocol allows that. Or user can write some queri like 'SELECT * FROM really_big_table WHERE $1 = $1` to force extended flow.

mp911de:
Right. I was heading more into the direction that if a `fetchSize` is declared, then `SimpleQueryPostgresqlStatement.execute()` would use `statementCache` and `ExtendedQueryMessageFlow` with an empty binding. By default, we would use `SimpleQueryMessageFlow`.

That way, we could provide a configuration property to configure a default fetch size to help applications that make regular use of cancellation with large data sets.

src/main/java/io/r2dbc/postgresql/client/ExtendedQueryMessageFlow.java
```diff
@@ -69,15 +76,55 @@ private ExtendedQueryMessageFlow() {
...
+            FluxSink<FrontendMessage> requestsSink = requestsProcessor.sink();
+            AtomicBoolean isCanceled = new AtomicBoolean(false);
+            return client.exchange(bindFlow.concatWith(requestsProcessor))
+                .handle((BackendMessage message, SynchronousSink<BackendMessage> sink) -> {
```
mp911de: Another thought crossed my mind: Right now, a query is required to fully complete before another one can be executed.

 What if we would break down cursored queries into smaller conversations that terminate either with `CommandComplete` or `PortalSuspended` messages? 

That way, we could interleave different queries and we would have smaller conversations in total. We could also factor in the current demand before requesting more data. I'm not sure how long a single portal can be kept open though.

Squiry: Error handling will be really tricky if we're going to allow multiple queries at one time. After server sends ErrorResponse it discards every request until it find Sync (that's why we're sending Sync after error). Also Sync will commit transaction (if there were now explicit transactions) and this will close every portal. We can have our connection in some really weird state after that: neverending conversations (since backend discarded requests) or closed portals (since backend closed them on Sync) are possible.

mp911de: Okay, I see that interleaving doesn’t make sense. I’m going to merge and polish the PR in the next days and take a look on fetch size with unparametrized statements.

Squiry: I was going to implement unparametrized statements fetch size in this PR somewhere near weekend, but ok, you can do it if you want to.

mp911de:
One more thing: Care to squash your commits into a single commit?

Squiry:
Sure, I've been keeping separate commits for review.

steven-sheehy:
Two questions:
1) If we set fetch size, is it a static request amount between client and server for all subsequent communication or is it just the initial request size and subsequent communication will adjust request based upon backpressure?
2) How can we set fetch size via DatabaseClient?

mp911de:
The fetch size remains constant for each cursor roundtrip. Right now, Spring Data's `DatabaseClient` does not have a mechanism to specify the fetch size or apply other customizations to `Statement`. That is going to be a future enhancement.

mp911de:
Thanks a lot. That's merged, polished, and backported now.

mp911de:
Related, I found a memory leak that was caused by closing the client while it was still receiving frames. The frames lingered in the `BackendMessageSubscriber.buffer`.

Issue #121, opened 2019-06-25
Send cancelation packet to Postgres server in order to interrupt running query 


OlegDokuka:
In some circumstances, it is wiser to send cancelation acknowledgment to the Postgres server (https://www.postgresql.org/docs/9.3/protocol-flow.html#AEN100004) to interrupt long running query rather than waiting for the complete response to drain from the connection.

Thus we have to consider to implement cancelation of subscription as a signal to Postgres over wired connection

PR #246, opened 2020-02-08
Canceling requests in progress with CancelRequest

Squiry:
#121


mp911de:

PR #257, opened 2020-03-07
Add default fetch size support

### LINKED ISSUE #256: Default fetch size
## Feature Request

Add a configuration option to set the default fetch size support on the ConnectionFactory.

There's currently no way to set the default fetch size globally or via higher level Spring Data operations like `@Query` or `DatabaseClient`. When fetch size is unset, queries are un-cursored and can't be cancelled.

Be able to set a default fetch size that is suitable for most queries. Individual statements can be overridden to have a custom value.

- Set fetch size via `@Query`
- Set fetch size via `DatabaseClient`
- Set fetch size on the statement

steven-sheehy:
Make sure that:

- [x] You have read the [contribution guidelines](https://github.com/r2dbc/.github/blob/master/CONTRIBUTING.adoc).
- [x] You use the code formatters provided [here](https://github.com/r2dbc/.github/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.

 

#### Issue description

Fixes #256
 
#### New Public APIs

- `PostgresqlConnectionConfiguration.builder().fetchSize()`
- `PostgresqlConnectionFactoryProvider.FETCH_SIZE`

#### Additional context

- Refactored some classes to pass and store `PostgresqlConnectionConfiguration` instead of duplicating the fetch size and forceBinary fields. Should make it more future proof to future additions
- I probably need more tests, but not sure where or what I should add so some guidance would be good here.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
@@ -25,7 +25,6 @@ import io.r2dbc.postgresql.codec.Codec;
 import io.r2dbc.postgresql.extension.CodecRegistrar;
 import io.r2dbc.postgresql.extension.Extension;
 import io.r2dbc.postgresql.util.Assert;
-import io.r2dbc.spi.Statement;
 import reactor.netty.tcp.SslProvider;
 import reactor.util.annotation.Nullable;
 
@@ -414,6 +413,7 @@ public final class PostgresqlConnectionConfiguration {
          * @return this {@code Builder}
          */
         public Builder fetchSize(int fetchSize) {
+            Assert.isTrue(fetchSize >= 0, "fetch size must be greater or equal zero");
             this.fetchSize = fetchSize;
             return this;
         }
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
@@ -37,6 +37,7 @@ import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.function.ToIntFunction;
 
 import static io.r2dbc.postgresql.message.frontend.Execute.NO_LIMIT;
 import static reactor.netty.tcp.SslProvider.DefaultConfigurationType.TCP;
@@ -61,7 +62,7 @@ public final class PostgresqlConnectionConfiguration {
 
     private final List<Extension> extensions;
 
-    private final int fetchSize;
+    private final ToIntFunction<String> fetchSize;
 
     private final boolean forceBinary;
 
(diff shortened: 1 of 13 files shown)


src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -60,6 +62,8 @@
 
     private final List<Extension> extensions;
 
+    private final int fetchSize;
```
mp911de: How about turning this into a `ToIntFunction<String>`?  The argument would be the actual SQL. That way, we can have a default fetchsize while being able to react to specific SQL statements to return a different fetch size.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionConfiguration.java
```diff
@@ -394,6 +406,18 @@ public Builder extendWith(Extension extension) {
...
 
+        /**
+         * Set the default number of rows to return when fetching results from a query instead deriving fetch size from
+         * back pressure. If the value specified is zero, then the hint is ignored.
+         *
+         * @param fetchSize the number of rows to fetch
+         * @return this {@code Builder}
+         */
+        public Builder fetchSize(int fetchSize) {
+            this.fetchSize = fetchSize;
```
mp911de: Lets add an assert that the number of rows is not negative.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -260,6 +265,11 @@ private static boolean convertToBoolean(Object value) {
             builder.port(port);
         }
 
+        Integer fetchSize = connectionFactoryOptions.getValue(FETCH_SIZE);
```
mp911de: Values can be either Integer or String as the URL parser cannot know the actual data type.

src/test/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProviderTest.java
```diff
@@ -152,6 +153,19 @@ void shouldCreateConnectionFactoryWithoutPassword() {
             .build())).isNotNull();
     }
 
+    @Test
+    void providerShouldConsiderFetchSize() {
```
mp911de: We should have also another test with providing the fetch size as string.

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -52,6 +52,11 @@
      */
     public static final Option<Boolean> AUTODETECT_EXTENSIONS = Option.valueOf("autodetectExtensions");
 
+    /**
+     * Fetch Size.
+     */
+    public static final Option<Integer> FETCH_SIZE = Option.valueOf("fetchSize");
```
mp911de: We should document the option in the readme.

steven-sheehy:
Thanks for the review. I think I've addressed all your feedback. Please take a look again.

mp911de:

PR #269, opened 2020-03-29
Add codec for hstore as Map (text and binary decode)

### LINKED ISSUE #268: Support hstore as default codec
## Feature Request

First of all, thank you for your driver !
I started a project recently and we chose to use r2dbc. My main domain is geographic services (OpenStreetMap data) and we often use PostgreSQL and Postgis.
In this domain, there is also a type we always use, it's `hstore`, and there are no codec for this type in r2dbc-postgresql.

`hstore` data type for storing sets of key/value pairs within a single PostgreSQL value and is not a default type in PostgreSQL, it's a module that need to be activated before the use. But it could still be a good addition.

I do not know if there is a large part of r2dbc users who use `hstore`, if they are few they can remain their responsibility (custom codec).
This is what I did for my project, I wrote my own Hstore codec and it works pretty well.

I want to share what I have done with the community, because if I needed it, maybe others will be in the same case. That's why I wrote the piece of code [here](https://github.com/Joxit/r2dbc-postgresql/tree/feat/hstore-support).

- The `hstore` id is `20832` (greater than `OID_CACHE_SIZE`)
- The Java type for `hstore` is `Map<String, String>`
- We must use raw types because we cannot have the class with  parameterized type (`Codec<Map>` instead of `Codec<Map<String, String>>`).

If this add makes sens, I can publish the PR.

Joxit:
#### Issue description

GitHub issue #268

#### Additional context

`hstore` OID is dynamics, I should add a test at connection with this query:
```sql
SELECT oid FROM pg_catalog.pg_type WHERE typname = 'hstore';
```

Closes #268

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactory.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactory.java
@@ -27,6 +27,7 @@ import io.r2dbc.postgresql.client.SSLConfig;
 import io.r2dbc.postgresql.client.SSLMode;
 import io.r2dbc.postgresql.client.StartupMessageFlow;
 import io.r2dbc.postgresql.codec.DefaultCodecs;
+import io.r2dbc.postgresql.codec.DynamicCodecs;
 import io.r2dbc.postgresql.extension.CodecRegistrar;
 import io.r2dbc.postgresql.message.backend.AuthenticationMessage;
 import io.r2dbc.postgresql.util.Assert;
@@ -202,6 +203,8 @@ public final class PostgresqlConnectionFactory implements ConnectionFactory {
 
         List<Publisher<?>> publishers = new ArrayList<>();
         publishers.add(setSchema(connection));
+
+        publishers.add(new DynamicCodecs().register(connection, byteBufAllocator, codecs));
 
         this.extensions.forEach(CodecRegistrar.class, it -> {
             publishers.add(it.register(connection, byteBufAllocator, codecs));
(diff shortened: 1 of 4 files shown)


Joxit:
My first idea for dynamic OIDs are this one, but it's not really elegant... This solution implies `HStoreCodec` class and constructor to be public....

```java
    private Publisher<?> prepareConnection(PostgresqlConnection connection, ByteBufAllocator byteBufAllocator, DefaultCodecs codecs) {

        List<Publisher<?>> publishers = new ArrayList<>();
        publishers.add(setSchema(connection));

        publishers.add(
            connection.createStatement("SELECT oid FROM pg_catalog.pg_type WHERE typname = 'hstore';")
                .execute()
                .flatMap(it -> it.map((row, rowMetadata) -> {
                        Integer hstoreOid = row.get("oid", Integer.class);
                        if (hstoreOid != null) {
                            codecs.addLast(new HStoreCodec(byteBufAllocator, hstoreOid));
                        }
                        return Mono.empty();
                    })
                )
        );

        this.extensions.forEach(CodecRegistrar.class, it -> {
            publishers.add(it.register(connection, byteBufAllocator, codecs));
        });

        return Flux.concat(publishers).then();
    }
```

The second solution could be a new class in `codec` package named `DynamicCodecRegistrar` which implement `CodecRegistrar` and will load all dynamic extensions (such as `hstore` and future ones ?)

Thought ?

Joxit:
Hi there, the PR is ready to review.
I added `DynamicCodecs` class. For now, it load only hstore, but will be able to load all built-in extensions.

mp911de:

PR #272, opened 2020-04-05
use "schema" as "search_path" connection option

### LINKED ISSUE #271: add support for providing "search_path" connection option
## Feature Request

Missing is a way to specify `search_path` in a R2DBC URL

In some cases, it's useful to be able to specify search_path in the URL, as it's a database specific aspect. 

I want to be able to specify connection URL as 

`spring.r2dbc.url=r2dbc:postgresql://localhost:5432/postgres?search_path=schema1,myschema`

An alternative (with Spring Data) is to have a bean implementing `ConnectionFactoryOptionsBuilderCustomizer` and make it do this: 

```
    @Override
    public void customize(ConnectionFactoryOptions.Builder builder) {
        if (StringUtils.isNotBlank(schemaSearchPath)) {
            builder.option(Option.valueOf("options"), Collections.singletonMap("search_path", schemaSearchPath));
        }
    }
```

This solution is sub-optimal as normally you would use same Spring config for different databases, and with H2 (for example) it will not work, because of different Option type and different option names. Would be better to be able to address all connection-specific aspects via single configuration aspect (R2DBC URL)

62mkv:
This change adds an alias `currentSchema` for `schema` configuration parameter; and uses this value as a value for "search_path" connection option (instead of executing `SET SCHEMA TO` command upon connection)

[resolves #271 ]

Make sure that:

- [x] You have read the [contribution guidelines](https://github.com/r2dbc/.github/blob/master/CONTRIBUTING.adoc).
- [ ] You use the code formatters provided [here](https://github.com/r2dbc/.github/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.

 

#### Issue description

 
Closes #271

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Mono<Connection> connectionMono = Mono.from(connectionFactory.create());
 | `forceBinary`     | Whether to force binary transfer.  Defaults to `false`. _(Optional)_
 | `preparedStatementCacheQueries` | Determine the number of queries that are cached in each connection. The default is `-1`, meaning there's no limit. The value of `0` disables the cache. Any other value specifies the cache size.
 | `options`         | A `Map<String, String>` of connection parameters. These are applied to each database connection created by the `ConnectionFactory`. Useful for setting generic [PostgreSQL connection parameters][psql-runtime-config]. _(Optional)_
-| `schema`          | The schema to set. _(Optional)_
+| `schema`          | The search path to set. _(Optional)_
 | `sslMode`         | SSL mode to use, see `SSLMode` enum. Supported values: `DISABLE`, `ALLOW`, `PREFER`, `REQUIRE`, `VERIFY_CA`, `VERIFY_FULL`. _(Optional)_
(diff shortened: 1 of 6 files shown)


62mkv:
hi @mp911de is to too bad for review? just to clarify: I've left one action (about formatters) unchecked, because when I applied (or thought to have applied) the recommended style and ran the Code Reformat action in IntelliJ, the code was re-formatted heavily so I had to manually adjust it.

It still feels that making something like

```
Map<String, String> options;
if (connectionFactoryOptions.getValue(OPTIONS)) instanceof String) {
   options = tryParseOptions(connectionFactoryOptions.getValue(OPTIONS));
} else {
   options = connectionFactoryOptions.getValue(OPTIONS);
}
```

might be a more generic way to address this (and similar) issues. What do you think ?

mp911de:
No worries. We weren't able yet to review this PR as we're busy with other projects.

mp911de:

PR #273, opened 2020-04-18
adds capability to specify connections options via URL

### LINKED ISSUE #271: add support for providing "search_path" connection option

62mkv:
added workaround for cases when "options" have been provided via URL and are represented with a String, not a Map)

Make sure that:

- [X] You have read the [contribution guidelines](https://github.com/r2dbc/.github/blob/master/CONTRIBUTING.adoc).
- [X] You use the code formatters provided [here](https://github.com/r2dbc/.github/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.

 

#### Issue description

Resolves #271

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
@@ -320,13 +320,12 @@ public final class PostgresqlConnectionFactoryProvider implements ConnectionFact
             builder.preparedStatementCacheQueries(convertToInt(preparedStatementCacheQueries));
         }
 
-        Map<String, String> options;
-        try {
-            options = connectionFactoryOptions.getValue(OPTIONS);
-        } catch (ClassCastException exception) {
-            Option<String> optionsAsString = Option.valueOf("options");
-            options = mapOptionsFromUrlString(connectionFactoryOptions.getValue(optionsAsString));
-        }
+        Option<Object> optionsAsObject = Option.valueOf("options");
+        Object optionsObject = connectionFactoryOptions.getValue(optionsAsObject);
+
+        Map<String, String> options = optionsObject instanceof String
+            ? mapOptionsFromUrlString((String) optionsObject)
+            : connectionFactoryOptions.getValue(OPTIONS);
 
         if (options != null) {
             builder.options(options);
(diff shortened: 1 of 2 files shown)


62mkv:
@mp911de can we please discuss this? It seems to address my original need in full and also unlocks many more potentially useful use-cases. 

Do you think it is something, that might be possible to progress? I could add tests as well, if you give me a hint where to put them and how should they look like...

src/main/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProvider.java
```diff
@@ -280,7 +281,14 @@ private static int convertToInt(Object value) {
             builder.preparedStatementCacheQueries(convertToInt(preparedStatementCacheQueries));
         }
 
-        Map<String, String> options = connectionFactoryOptions.getValue(OPTIONS);
+        Map<String, String> options;
+        try {
+            options = connectionFactoryOptions.getValue(OPTIONS);
+        } catch (ClassCastException exception) {
```
mp911de, line 287: This should be rather `Object options = getValue(…)` and then inspecting whether the options are a string or a map.

62mkv: sure, will update and rebase

62mkv:
@mp911de requested changes seem to be addressed

mp911de:

PR #298, opened 2020-07-10
Support for "time with time zone" type

### LINKED ISSUE #294: Support for "time with time zone" type
When you are working with time, you often need a timezone.  Right now I have to save time with timezone as string to db, so I loose option for native query in those columns, that affects my service performance

Add mapping time with time zone to OffsetTime

none

anshlykov:
Make sure that:

#### Issue description

Resolves #294

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/OffsetTimeCodec.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/OffsetTimeCodec.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


src/main/java/io/r2dbc/postgresql/codec/OffsetTimeCodec.java
```diff
@@ -0,0 +1,65 @@
...
+        Assert.requireNonNull(buffer, "byteBuf must not be null");
+
+        return decodeTemporal(buffer, dataType, format, OffsetTime.class, OffsetTime::from);
+    }
+
+    @Override
+    Parameter doEncode(OffsetTime value) {
+        Assert.requireNonNull(value, "value must not be null");
+
+        return create(TIMETZ, FORMAT_TEXT, () -> ByteBufUtils.encode(this.byteBufAllocator, value.toString()));
```
mp911de, line 57: Would it make sense to encode `OffsetTime` always binary to preserve `offsetSec`?

davecramer: hmmmm... I'm not sure this comment is appropriate as I don't exactly know what FORMAT_TEXT means. Assuming it means the way the data is transmitted. If you go to binary there will be no offset. Postgres does not store offsets in the column. All data is stored in UTC and then the offset is whatever the client sets it to

mp911de: Text is the way data is represented on the wire. In this case, please disregard my earlier comment. I somehow lived under the impression that binary had more details than the text representation.

davecramer: Ironically it has less :)

anshlykov:
@mp911de  I added the test to `AbstractCodecIntegrationTests`

mp911de:

PR #305, opened 2020-07-25
feature: adding support for lseg type

avinash-anand:
Make sure that:

- [X] You have read the [contribution guidelines](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/.github/CONTRIBUTING.adoc).
- [X] You have created a feature request first to discuss your contribution intent. Please reference the feature request ticket number in the pull request.
- [X] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.
- [X] You submit test cases (unit or integration tests) that back your changes.

 

#### Issue description

related to issue previously opened support for geometric types - #282

PR #306, opened 2020-07-27
Support for geospatial types - point, line, box, polygon etc

### LINKED ISSUE #282: Support for geospatial types - point, line, box, polygon etc
## Feature Request
Please don't mind as I'm a beginner wrt requesting OSS features.
I would like to have support for geo-spatial types like point, line, polygon etc.

Our current project is JDBC based and there we use postgis-jdbc to save our data along with geospatial data in postgresql. we save mostly coordinates (Points) or lines or polygons for some markings done on maps. We were wondering if we can use r2dbc, if it had the support for these data types.

I would like to have support for geospatial types of postgresql.

#### Geometry type support

- [x] `point` (#283)
- [ ] `line`
- [ ] `lseg`
- [ ] `box`
- [ ] `path` (open, closed)
- [ ] `polygon`
- [x] `circle` (#300)

anshlykov:
Resolves #282

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/Line.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/Line.java
@@ -1,0 +1,62 @@
+package io.r2dbc.postgresql.codec;
+
+import java.util.Objects;
+
+/**
+ * Value object that maps to the {@code line} datatype in Postgres.
+ * <p>
+ * Uses {@code double} to represent the coordinates.
+ */
+public final class Line {
+
+    private final double a;
+    private final double b;
+    private final double c;
+
+    private Line(double a, double b, double c) {
+        this.a = a;
+        this.b = b;
+        this.c = c;
+    }
+
+    public static Line of(double a, double b, double c) {
+        return new Line(a, b, c);
+    }
+
+    public double getA() {
+        return this.a;
+    }
+
+    public double getB() {
+        return this.b;
+    }
+
+    public double getC() {
+        return this.c;
+    }
+
+    @Override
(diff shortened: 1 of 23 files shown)


mp911de:
Thanks for your contribution. Please note there's another pull request (#305) that ships with `lseg` support. Since this PR covers the remaining geo codecs it makes sense to consider this one over #305.

src/main/java/io/r2dbc/postgresql/codec/Line.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Objects;
+
+/**
+ * Value object that maps to the {@code line} datatype in Postgres.
+ * <p>
+ * Uses {@code double} to represent the coordinates.
+ */
+public final class Line {
+
+    private final double a;
```
mp911de, line 12: We should add a bit of Javadoc to the fields.

src/main/java/io/r2dbc/postgresql/codec/Line.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final double c;
+
+    private Line(double a, double b, double c) {
+        this.a = a;
+        this.b = b;
+        this.c = c;
+    }
+
+    public static Line of(double a, double b, double c) {
```
mp911de, line 24: Would it make sense to add factory methods accepting two points or even x1/y1, x2/y2?

anshlykov:
@mp911de thanks for the review. I updated the `Line` class

mp911de:

PR #309, opened 2020-07-30
added parameter bind logging

### LINKED ISSUE #278: add capability to log query parameters (bindings) values

I'd like to be able to specify logging level (say, `io.r2dbc.postgresql.PARAM`) which, when set to DEBUG or TRACE, would enable logging of parameters. 

`r2bc-proxy` is an alternative, but seems to be a big overhead for such a simple use-case 

A mention in the README should suffice as a documentation

62mkv:
closes #278

Make sure that:

Not sure what tests should I provide along with this? 

 

#### Issue description

#278 
 
#### New Public APIs

none

#### Additional context

Query Parameter values are now visible in the logs

### DIFF
--- a/README.md
+++ b/README.md
@@ -478,6 +478,7 @@ Logging facilities:
 
 * Driver Logging (`io.r2dbc.postgresql`)
 * Query Logging (`io.r2dbc.postgresql.QUERY` on `DEBUG` level)
+* Parameters' values Logging (`io.r2dbc.postgresql.PARAM` on `DEBUG` level)
 * Transport Logging (`io.r2dbc.postgresql.client`)
     * `DEBUG` enables `Message` exchange logging
     * `TRACE` enables traffic logging
--- a/src/main/java/io/r2dbc/postgresql/BindingLogger.java
+++ b/src/main/java/io/r2dbc/postgresql/BindingLogger.java
@@ -1,0 +1,20 @@
+package io.r2dbc.postgresql;
+
+import reactor.util.Logger;
+import reactor.util.Loggers;
+
+/**
+ * Binding logger to log parameter bindings
+ */
+final class BindingLogger {
+
+    private static final Logger BINDING_LOGGER = Loggers.getLogger("io.r2dbc.postgresql.PARAM");
+
+    static void logBinding(int index, String bindValue) {
+        BINDING_LOGGER.debug("Parameter {} value bound: {}", index, bindValue);
+    }
+
+    static void logBinding(String name, String bindValue) {
+        BINDING_LOGGER.debug("Parameter {} value bound: {}", name, bindValue);
+    }
+}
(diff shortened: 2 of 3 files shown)


mp911de:
Thanks for submitting a pull request. Care to attach a snippet as comment how the logs would look like?

I'm considering to slightly tweak the logging behavior. Since we have already a query logger, we might want to keep the actual SQL and the parameters somewhat close together so one can reason about the query and its parameters.

Let's see how the log output would look like before we continue here.

62mkv:
Hi. This is how the output looks like for me:

```
Started
2020-08-11 17:04:06.741 DEBUG 8288 --- [  restartedMain] io.r2dbc.postgresql.PARAM                : Parameter 0 value bound: %a%
2020-08-11 17:04:06.744 DEBUG 8288 --- [  restartedMain] io.r2dbc.postgresql.PARAM                : Parameter 1 value bound: 1
2020-08-11 17:04:06.744 DEBUG 8288 --- [  restartedMain] io.r2dbc.postgresql.PARAM                : Parameter 2 value bound: 2020-01-01
2020-08-11 17:04:06.773 DEBUG 8288 --- [actor-tcp-nio-1] io.r2dbc.postgresql.QUERY                : Executing query: SELECT TEMPDATA.id, TEMPDATA.textdata, TEMPDATA.tzdata FROM TEMPDATA WHERE TEMPDATA.textdata LIKE $1 AND (TEMPDATA.id > $2) AND (TEMPDATA.tzdata > $3)
Finished
```

> I'm considering to slightly tweak the logging behavior. Since we have already a query logger, we might want to keep the actual SQL and the parameters somewhat close together so one can reason about the query and its parameters.

Could you please provide a little bit more details, on how you see it accomplished? 

Thanks!

62mkv:
where I see a problem with the above (apart from off-by-one numbering) is what might happen in mass-parallel scenarios. So arguably, if we could put some unique query execution id into both QUERY and PARAM logger output so user could correlate those together, would be great. However at the moment I have no idea how to achieve this..

Squiry:
I believe we need some connection id in log, in every log, not only bindings, that would help a lot.

ttddyy:
FYI, If you want to log bindings and query all together(also connection id), r2dbc-proxy provides such capability.
[Here is sample log.](https://github.com/r2dbc/r2dbc-proxy#query-logging)
(It is multilined for display purpose but actual logging is single line, and configurable to own format)

mp911de:
The log output looks good already. I second @Squiry's view, that we need to add context. Single-process logs are easy to understand, but in production environments we get way more log output than that.

In the SQL Server driver, we've introduced [`ConnectionContext`](https://github.com/r2dbc/r2dbc-mssql/blob/bb52fe64f2b43b605e71648a8774f48d2ff33659/src/main/java/io/r2dbc/mssql/client/ConnectionContext.java#L35) that is a factory for log messages holding the connection/channel Id for easier tracing. We could do here the same.

I'll be off for around two weeks, I can have a look afterward. Alternatively, feel free to check out the SQL server driver and introduce the contextual log message factory.

Squiry:
We have a [processId](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java#L119) that can be used as a connection id provided by backend, but we have it only _after_ connection succeed.

mp911de:
I filed #339 to capture more context so we can correlate bind calls to the actual connection.

mp911de:

62mkv:
Thanks @mp911de !

PR #310, opened 2020-07-31
Add priority to EnumCodec registrar

### LINKED ISSUE #303: Add priority to EnumCodec registrar
Right now, `EnumCodecs` are registered last in the codec chain. We should add a configuration option to `EnumCodec.Builder` to allow configuring whether enum codecs should be registered first or last.

anshlykov:
#### Issue description

Resolves #303 
 
#### New Public APIs

- `EnumCodec.Builder#withEnum(String, Class<? extends Enum<?>>, RegistrationPriority)`
- enum `EnumCodec.Builder.RegistrationPriority`. Maybe it make sense to move this inner enum to upper level.

#### Additional context

- Added `MockPostgresqlConnection`, `MockPostgresqlResult` and `MockPostgresqlStatement` for tests. These classes are partially implemented, but this is sufficient for test cases in the PR.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/EnumCodec.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/EnumCodec.java
@@ -119,7 +119,9 @@ public final class EnumCodec<T extends Enum<T>> implements Codec<T> {
      */
     public static final class Builder {
 
-        private final Map<String, EnumMappingConfig> mapping = new LinkedHashMap<>();
+        private final Map<String, Class<? extends Enum<?>>> mapping = new LinkedHashMap<>();
+
+        private RegistrationPriority registrationPriority = RegistrationPriority.LAST;
 
         /**
          * Add a Postgres enum type to {@link Enum} mapping.
@@ -129,18 +131,6 @@ public final class EnumCodec<T extends Enum<T>> implements Codec<T> {
          * @return this {@link Builder}
          */
         public Builder withEnum(String name, Class<? extends Enum<?>> enumClass) {
-            return withEnum(name, enumClass, RegistrationPriority.LAST);
-        }
-
-        /**
-         * Add a Postgres enum type to {@link Enum} mapping.
-         *
-         * @param name      name of the Postgres enum type
-         * @param enumClass the corresponding Java type
-         * @param priority  the codec registration priority
-         * @return this {@link Builder}
-         */
-        public Builder withEnum(String name, Class<? extends Enum<?>> enumClass, RegistrationPriority priority) {
             Assert.requireNotEmpty(name, "Postgres type name must not be null");
             Assert.requireNonNull(enumClass, "Enum class must not be null");
             Assert.isTrue(enumClass.isEnum(), String.format("Enum class %s must be an enum type", enumClass.getName()));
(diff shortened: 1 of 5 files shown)


src/main/java/io/r2dbc/postgresql/codec/EnumCodec.java
```diff
@@ -129,18 +129,31 @@ public Parameter encodeNull() {
...
+
+        /**
+         * Add a Postgres enum type to {@link Enum} mapping.
+         *
+         * @param name      name of the Postgres enum type
+         * @param enumClass the corresponding Java type
+         * @param priority  the codec registration priority
+         * @return this {@link Builder}
+         */
+        public Builder withEnum(String name, Class<? extends Enum<?>> enumClass, RegistrationPriority priority) {
```
mp911de, line 143: Let's keep it for the time being more simple. We probably don't need to control priority on a per-codec basis, it's rather whether all enum codecs should come first/last.

That being said, we don't need to expose `RegistrationPriority` and we can keep `withEnum(…)` signatures as-is.

src/main/java/io/r2dbc/postgresql/codec/EnumCodec.java
```diff
@@ -181,6 +198,27 @@ public CodecRegistrar build() {
...
 
+        /**
+         * An enumeration of codec registration priorities.
+         */
+        public enum RegistrationPriority {
+            FIRST,
+            LAST
+        }
+
+        private static final class EnumMappingConfig {
```
mp911de, line 209: Following the comment earlier, we don't need this type.

src/test/java/io/r2dbc/postgresql/codec/EnumCodecUnitTests.java
```diff
@@ -40,6 +52,70 @@ void shouldRejectMultipleMappingForTypeName() {
...
+                    .identified("typcategory", String.class, "E")
+                    .build())
+                .build())
+            .build();
+        MockPostgresqlConnection mockPostgresqlConnection = new MockPostgresqlConnection(mockPostgresqlStatement);
+
+        Publisher<Void> register = codecRegistrar.register(mockPostgresqlConnection, mockByteBufAllocator, mockCodecRegistry);
+        StepVerifier.create(register).verifyComplete();
+
+        Mockito.verify(mockCodecRegistry, Mockito.only()).addFirst(Mockito.any(EnumCodec.class));
```
mp911de, line 80: Care to switch to static imports for Mockito?

mp911de:

src/test/java/io/r2dbc/postgresql/codec/EnumCodecUnitTests.java
anshlykov: done

src/main/java/io/r2dbc/postgresql/codec/EnumCodec.java
anshlykov: The class has deleted

src/main/java/io/r2dbc/postgresql/codec/EnumCodec.java
anshlykov: The method `withEnum(String, Class<? extends Enum<?>>, RegistrationPriority)` has deleted

PR #318, opened 2020-08-17
Accept SSL certificates by providing a URL to use cert from within a jar

### LINKED ISSUE #313: Accept SSL certificates by providing a URL to use cert from within a jar
I have a `cert.pem` in my `src/main/resources`
I need it when running the app jarred as well as unjarred.

Specifying it faithfully in my `application.properties` as `spring.r2dbc.properties.sslRootCert=classpath:cert.pem` fails as r2dbc does not load the file via classpath: https://github.com/pgjdbc/r2dbc-postgresql/blob/main/src/main/java/io/r2dbc/postgresql/util/Assert.java#L130

```
Caused by: java.lang.IllegalArgumentException: sslRootCert must not be null and must exist
	at io.r2dbc.postgresql.util.Assert.requireFileExistsOrNull(Assert.java:131) ~[r2dbc-postgresql-0.8.4.RELEASE.jar:0.8.4.RELEASE]
	at io.r2dbc.postgresql.PostgresqlConnectionConfiguration$Builder.sslRootCert(PostgresqlConnectionConfiguration.java:641) ~[r2dbc-postgresql-0.8.4.RELEASE.jar:0.8.4.RELEASE]
```

I see two options: Either load the specified File via classloader#getResourceAsStream or accept an URL as a cert parameter instead of a String.

The issue was originally raised here: https://github.com/spring-projects/spring-data-r2dbc/issues/430

Isabek:
[resolves #313]

Make sure that:

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/util/Assert.java
+++ b/src/main/java/io/r2dbc/postgresql/util/Assert.java
@@ -19,6 +19,7 @@ package io.r2dbc.postgresql.util;
 import reactor.util.annotation.Nullable;
 
 import java.io.File;
+import java.net.URL;
 
 /**
  * Assertion library for the implementation.
@@ -145,6 +146,25 @@ public final class Assert {
         if (!expression) {
             throw new IllegalArgumentException(message);
         }
+    }
+
+    /**
+     * Checks that the provided URL exists or null.
+     *
+     * @param url    url to check
+     * @param message the message to use in exception if type is not as required
+     * @return existing url
+     * @throws IllegalArgumentException if {@code value} is not of the required type
+     * @throws IllegalArgumentException if {@code value}, {@code type}, or {@code message} is {@code null}
+     */
+    public static URL requireUrlExistsOrNull(@Nullable URL url, String message) {
+        if (url == null) {
+            throw new IllegalArgumentException(message);
+        }
+        if (!new File(url.getFile()).exists()) {
+            throw new IllegalArgumentException(message);
+        }
+        return url;
     }
 
 }
--- a/src/main/java/io/r2dbc/postgresql/util/Assert.java
+++ b/src/main/java/io/r2dbc/postgresql/util/Assert.java
@@ -19,12 +19,15 @@ package io.r2dbc.postgresql.util;
 import reactor.util.annotation.Nullable;
 
 import java.io.File;
+import java.net.MalformedURLException;
 import java.net.URL;
 
 /**
  * Assertion library for the implementation.
  */
 public final class Assert {
+
+    private static final String CLASSPATH_PREFIX = "classpath:";
 
     private Assert() {
     }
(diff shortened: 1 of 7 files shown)


Squiry:
@Isabek your changes need to be covered by tests
1. Integration test with classpath url [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/src/test/java/io/r2dbc/postgresql/client/ReactorNettyClientIntegrationTests.java#L674).
2. Test cases for configuration parsing and mapping string values to URLs [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/src/test/java/io/r2dbc/postgresql/PostgresqlConnectionFactoryProviderUnitTests.java).
3. Test cases for possible mistakes in configuration (like file/resource doesn't exist) [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/src/test/java/io/r2dbc/postgresql/PostgresqlConnectionConfigurationUnitTests.java).

mp911de:
`classpath:` is a Spring-specific feature hence it's not applicable in our context. Please use `ClassLoader.getResource(…)` to construct an URL pointing to class-path resources.

mp911de:
Any update on this pull request? If not, we will close this one in a few days.

Isabek:
@mp911de sorry. I will update it on weekends.

Isabek:
@mp911de I have pushed my last changes. Could you review it, please?

davecramer:
@Isabek can we also update the README to reflect this option ?

src/main/java/io/r2dbc/postgresql/util/Assert.java
```diff
@@ -147,4 +132,44 @@ public static void isTrue(boolean expression, String message) {
...
+        if (path == null) {
+            throw new IllegalArgumentException(message);
+        }
+
+        if (path.startsWith(CLASSPATH_PREFIX)) {
+            path = path.substring(CLASSPATH_PREFIX.length());
+        }
+
+        ClassLoader classLoader = Assert.class.getClassLoader();
+        URL url = classLoader != null ? classLoader.getResource(path) : ClassLoader.getSystemResource(path);
```
mp911de, line 154: How is the classloader supposed to be null?

src/main/java/io/r2dbc/postgresql/util/Assert.java
```diff
@@ -19,12 +19,16 @@
...
 import java.io.File;
+import java.net.MalformedURLException;
+import java.net.URL;
 
 /**
  * Assertion library for the implementation.
  */
 public final class Assert {
 
+    private static final String CLASSPATH_PREFIX = "classpath:";
```
mp911de, line 30: That isn't a generally supported URL scheme. It only works because Spring is on the class path.

src/main/java/io/r2dbc/postgresql/util/Assert.java
```diff
@@ -147,4 +132,44 @@ public static void isTrue(boolean expression, String message) {
...
+    /**
+     * Checks that the provided URL exists or null.
+     *
+     * @param path    path to check
+     * @param message the message to use in exception if type is not as required
+     * @return existing url
+     * @throws IllegalArgumentException if {@code value} is not of the required type
+     * @throws IllegalArgumentException if {@code value}, {@code type}, or {@code message} is {@code null}
+     */
+    public static URL requireUrlExistsOrNull(@Nullable String path, String message) {
```
mp911de, line 144: We should revert changes to the `Assert` class and instead, introduce rather an utility method to figure out, whether the given `ConnectionFactoryOption` tries to point to a real file or to a URL. Checking for a scheme would be a good approach if the configured string doesn't point to a file.

We might still run into an ambiguity on windows (`c:\foo\bar.crt` vs. `classpath:foo.crt` or `classpath:/foo`). Not sure how to resolve that.

mp911de (changes requested):

mp911de:
@Isabek I'm happy to take this PR from here as it requires a bit of experimentation. WDYT?

Isabek:
> @Isabek I'm happy to take this PR from here as it requires a bit of experimentation. WDYT?

Sure.

mp911de:

PR #321, opened 2020-08-19
Parse UTC timestamps with +00:00 offset

### LINKED ISSUE #320: Issue parsing UTC timestamptz from CockroachDB

I want to use this driver to connect to [CockroachDB](https://github.com/cockroachdb/cockroach) (a distributed database which uses the PostgreSQL wire protocol). I encountered an error when trying to parse a UTC timestamp. Here's a quick repro case, running with a CockroachDB backend.

```
		Map<String, String> options = new HashMap<>();
		options.put("timezone", "UTC");

		PostgresqlConnectionFactory connectionFactory = new PostgresqlConnectionFactory(PostgresqlConnectionConfiguration.builder()
				.host("localhost")
				.port(26257)
				.username("root")
				.database("defaultdb")
				.options(options)
				.build());

		Mono<Connection> mono = Mono.from(connectionFactory.create());
		
		Flux<OffsetDateTime> now = mono.flatMapMany(connection -> connection
				.createStatement("SELECT now() as now")
				.execute())
				.flatMap(it -> it.map((row, rowMetadata) -> row.get("now", OffsetDateTime.class)));
```

This code produces an error like:
```
java.time.format.DateTimeParseException: Text '2020-08-18 21:58:26.549368+00:00' could not be parsed, unparsed text found at index 29
	at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1952)
	at java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1851)
	at io.r2dbc.postgresql.codec.AbstractTemporalCodec.decodeTemporal(AbstractTemporalCodec.java:135)
	at io.r2dbc.postgresql.codec.AbstractTemporalCodec.decodeTemporal(AbstractTemporalCodec.java:95)
```

The error happens because CockroachDB returns UTC timestamps with an offset of `+00:00`. This differs from Postgres, which uses an offset of `+00`. The [DateTimeFormatter used by r2dbc](https://github.com/pgjdbc/r2dbc-postgresql/blob/440b45dd5cfd899fc911702bf081bbf1e5843cea/src/main/java/io/r2dbc/postgresql/codec/PostgresqlDateTimeFormatter.java#L51) is a bit restrictive about how it parses UTC timestamps. As the [docs for DateTimeFormatterBuilder](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatterBuilder.html#appendOffset-java.lang.String-java.lang.String-) describe:
> The "no offset" text controls what text is printed when the total amount of the offset fields to be output is zero. Example values would be 'Z', '+00:00', 'UTC' or 'GMT'. Three formats are accepted for parsing UTC - the "no offset" text, and the plus and minus versions of zero defined by the pattern.

The following change in [PostgresqlDateTimeFormatter](https://github.com/pgjdbc/r2dbc-postgresql/blob/440b45dd5cfd899fc911702bf081bbf1e5843cea/src/main/java/io/r2dbc/postgresql/codec/PostgresqlDateTimeFormatter.java) addresses the issue. The change works with both PostgreSQL and CockroachDB.

```
-            .appendOffset("+HH:mm", "+00")
+            .appendOffset("+HH:mm", "+00:00")
```

A patch could be made to CockroachDB to make it use a `+00` offset for UTC timestamps. This would take a while to get released.

There isn't anything to document.

rafiss:
Make sure that:

#### Issue description

Previously, the timestamp parser would throw an exception if it received
a value with an offset of "+00:00". It would only work if the offset
were "+00". Now, it works with both formats.

Fixes #320
 
#### New Public APIs

There are no new APIs.

#### Additional context

I encountered this issue when using this driver to connect to CockroachDB.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/PostgresqlDateTimeFormatter.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/PostgresqlDateTimeFormatter.java
@@ -48,7 +48,7 @@ class PostgresqlDateTimeFormatter {
             .appendFraction(NANO_OF_SECOND, 0, 9, true)
             .optionalEnd()
             .optionalStart()
-            .appendOffset("+HH:mm", "+00")
+            .appendOffset("+HH:mm", "+00:00")
             .optionalEnd()
             .toFormatter();
 
--- a/src/test/java/io/r2dbc/postgresql/codec/InstantCodecUnitTests.java
+++ b/src/test/java/io/r2dbc/postgresql/codec/InstantCodecUnitTests.java
@@ -61,8 +61,21 @@ final class InstantCodecUnitTests {
         ZonedDateTime zonedDateTime = ZonedDateTime.parse("2018-11-05T00:20:25.039883+09:00[Asia/Tokyo]");
         Instant instant = zonedDateTime.toInstant();
 
+        assertThat(new InstantCodec(TEST).decode(encode(TEST, "2018-11-05 00:20:25.039883+09:00"), TIMESTAMPTZ.getObjectId(), FORMAT_TEXT, Instant.class))
+                .isEqualTo(instant);
         assertThat(new InstantCodec(TEST).decode(encode(TEST, "2018-11-05 00:20:25.039883+09"), TIMESTAMPTZ.getObjectId(), FORMAT_TEXT, Instant.class))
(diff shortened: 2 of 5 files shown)


mp911de:

PR #338, opened 2020-10-14
Use RoundRobinInetAddressResolver for DNS resolution

johnny:
#### Issue description

As an alternative to #120 this solution will allow the client to connect to a postgres master/slave setup through multiple pgbouncers.

#### Considerations

This is impossible to unit test properly, but I can provide credentials to manually test the behavior of the resolver.

This change should be backwards compatible, so I didn't make the resolving behavior configurable.

#### Which Version?
Should this feature be merged to main or also to the 0.8.x branch?

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/client/BalancedResolverGroup.java
+++ b/src/main/java/io/r2dbc/postgresql/client/BalancedResolverGroup.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.r2dbc.postgresql.client;
+
+import io.netty.resolver.AddressResolver;
+import io.netty.resolver.AddressResolverGroup;
+import io.netty.resolver.DefaultNameResolver;
+import io.netty.resolver.RoundRobinInetAddressResolver;
+import io.netty.util.concurrent.EventExecutor;
+
+import java.net.InetSocketAddress;
+
+/**
+ * When the {@link InetSocketAddress} resolves to multiple IP addresses, pick one randomly.
+ *
+ * @since 0.8.6
+ */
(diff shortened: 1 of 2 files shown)


src/main/java/io/r2dbc/postgresql/client/BalancedResolverGroup.java
```diff
@@ -0,0 +1,21 @@
+package io.r2dbc.postgresql.client;
```
mp911de, line 1: Nit: Please add a license header (see other files) with 2020 as the inception year.

src/main/java/io/r2dbc/postgresql/client/BalancedResolverGroup.java
```diff
@@ -0,0 +1,21 @@
...
+import io.netty.resolver.AddressResolver;
+import io.netty.resolver.AddressResolverGroup;
+import io.netty.resolver.DefaultNameResolver;
+import io.netty.resolver.RoundRobinInetAddressResolver;
+import io.netty.util.concurrent.EventExecutor;
+
+import java.net.InetSocketAddress;
+
+/**
+ * When the {@link InetSocketAddress} resolves to multiple IP addresses, pick one randomly.
```
mp911de, line 12: Nit: Add `@since 0.8.6` tag

src/main/java/io/r2dbc/postgresql/client/BalancedResolverGroup.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public class BalancedResolverGroup extends AddressResolverGroup<InetSocketAddress> {
```
mp911de, line 14: No need to have this class `public`. Please make it package-private and `final`. Since there's no state associated, it would make sense to keep a static instance (singleton) instead of creating a new instance on each connect.

src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
```diff
@@ -360,7 +360,9 @@ private void handleParameterStatus(ParameterStatus message) {
         Assert.requireNonNull(socketAddress, "socketAddress must not be null");
         Assert.requireNonNull(settings, "settings must not be null");
 
-        TcpClient tcpClient = TcpClient.create(settings.getConnectionProvider()).remoteAddress(() -> socketAddress);
+        TcpClient tcpClient = TcpClient.create(settings.getConnectionProvider())
```
mp911de, line 363: Let's apply customization only if the `SocketAddress` is a `InetSocketAddress`. We don't need round-robin for unix domain sockets.

mp911de (commented):
Looks good to me. I added comments about a few minor issues.

src/main/java/io/r2dbc/postgresql/client/BalancedResolverGroup.java
johnny: done

johnny:
Thanks for the review. All comments should be addressed.

mp911de:

Issue #361, opened 2020-12-06
Unable to decode array_agg(varchar) type.


sumitsum:
## Bug Report
postgresql r2dbc driver throws exception when trying to read array_agg(varchar) type from postgresql. It throws the following exception:
`java.lang.IllegalArgumentException: Cannot decode value of type java.lang.Object `

#### Versions
- Driver: 0.8.6.RELEASE 
- Database: postgresql
- Java: openjdk 11.0.9
- OS: MacOS 10.15.7 Catalina

#### Current Behavior
r2dbc driver throws exception on reading array_agg(varchar) type value. The type returned is "`null`". Exception gets thrown when doing `row.get(col_name);`

<details>

```
java.lang.AssertionError: expectation "assertNext" failed (expected: onNext(); actual: onError(java.lang.IllegalArgumentException: Cannot decode value of type java.lang.Object))

	at reactor.test.ErrorFormatter.assertionError(ErrorFormatter.java:105)
	at reactor.test.ErrorFormatter.failPrefix(ErrorFormatter.java:94)
	at reactor.test.ErrorFormatter.fail(ErrorFormatter.java:64)
	at reactor.test.ErrorFormatter.failOptional(ErrorFormatter.java:79)
	at reactor.test.DefaultStepVerifierBuilder.lambda$consumeNextWith$1(DefaultStepVerifierBuilder.java:256)
	at reactor.test.DefaultStepVerifierBuilder$SignalEvent.test(DefaultStepVerifierBuilder.java:2112)
	at reactor.test.DefaultStepVerifierBuilder$DefaultVerifySubscriber.onSignal(DefaultStepVerifierBuilder.java:1408)
	at reactor.test.DefaultStepVerifierBuilder$DefaultVerifySubscriber.onExpectation(DefaultStepVerifierBuilder.java:1356)
	at reactor.test.DefaultStepVerifierBuilder$DefaultVerifySubscriber.onError(DefaultStepVerifierBuilder.java:1030)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.secondError(MonoFlatMap.java:185)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onError(MonoFlatMap.java:251)
	at reactor.core.publisher.MonoSubscribeOn$SubscribeOnSubscriber.onError(MonoSubscribeOn.java:150)
[...]

PR #362, opened 2020-12-07
Add NAME type to StringArray codec

davecramer:
@mp911de can you point me in the right direction here. expectNext doesn't expect the right thing?
attempt to fix issue #361 reading array_agg

### DIFF
--- a/src/test/java/io/r2dbc/postgresql/ArrayAggIntegrationTests.java
+++ b/src/test/java/io/r2dbc/postgresql/ArrayAggIntegrationTests.java
@@ -2,49 +2,42 @@ package io.r2dbc.postgresql;
 
 import io.r2dbc.postgresql.api.PostgresqlConnection;
 import io.r2dbc.postgresql.api.PostgresqlResult;
-import io.r2dbc.postgresql.codec.Json;
-import io.r2dbc.postgresql.util.PostgresqlServerExtension;
+import io.r2dbc.spi.ColumnMetadata;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.RegisterExtension;
 import org.springframework.jdbc.core.JdbcOperations;
+import static org.assertj.core.api.Assertions.assertThat;
 import reactor.test.StepVerifier;
 
-public class ArrayAggIntegrationTests {
-  @RegisterExtension
-  static final PostgresqlServerExtension SERVER = new PostgresqlServerExtension();
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
-  private final PostgresqlConnectionConfiguration configuration = PostgresqlConnectionConfiguration.builder()
-      .database(SERVER.getDatabase())
-      .host(SERVER.getHost())
-      .port(SERVER.getPort())
-      .password(SERVER.getPassword())
-      .username(SERVER.getUsername())
-      .forceBinary(true)
-      .build();
+public class ArrayAggIntegrationTests extends AbstractIntegrationTests {
 
-  private final PostgresqlConnectionFactory connectionFactory = new PostgresqlConnectionFactory(this.configuration);
 
   @Test
   void shouldReadArrayAgg() {
 
     JdbcOperations jdbcOperations = SERVER.getJdbcOperations();
-    jdbcOperations.execute("DROP TABLE IF EXISTS my_table;");
-    jdbcOperations.execute("CREATE TABLE my_table (t text)");
+    jdbcOperations.execute("DROP TABLE IF EXISTS name_table;");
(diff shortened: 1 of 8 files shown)


mp911de:
I think the latest Docker version on Mac has changed something that requires a version bump to Testcontainers 1.15.

davecramer:
Actually the title of this is a bit mis-leading. I should break this into two PR's. I had to update testcontainers to test this on my mac. The real issue is how to fix the array_arg stuff

src/test/java/io/r2dbc/postgresql/ArrayAggIntegrationTests.java
```diff
@@ -0,0 +1,52 @@
...
+        .verifyComplete();
+
+    connection.createStatement("SELECT array_agg(t) FROM my_table")
+        .execute()
+        .flatMap(it -> it.map((row, rowMetadata) -> row.get("array_agg")))
+        .as(StepVerifier::create)
+        .expectNext(new Object[]{"hello", "world", null})
+        .verifyComplete();
+
+    connection.close().block();
```
davecramer, line 50: @mp911de I'm not sure how to do this bit?

src/test/java/io/r2dbc/postgresql/ArrayAggIntegrationTests.java
```diff
@@ -0,0 +1,52 @@
...
+import io.r2dbc.postgresql.api.PostgresqlConnection;
+import io.r2dbc.postgresql.api.PostgresqlResult;
+import io.r2dbc.postgresql.codec.Json;
+import io.r2dbc.postgresql.util.PostgresqlServerExtension;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.springframework.jdbc.core.JdbcOperations;
+import reactor.test.StepVerifier;
+
+public class ArrayAggIntegrationTests {
```
mp911de, line 12: You can find the formatter config at `intellij-style.xml`

src/test/java/io/r2dbc/postgresql/ArrayAggIntegrationTests.java
mp911de: You can subclass `AbstractIntegrationTests`. That will give you access to a `connection` field so you don't need to worry about the connection lifecycle. Other than that, `close.block()` (or `close().as(StepVerifier::create).verifyComplete()`) are fine.

mp911de:
Do you want to split the PR or should I take care of the dependency upgrade and you simply rebase your PR?

davecramer:
Do the latter. Update the dependency. It's irrelevant to my PR. Thanks

mp911de:
Upgraded to Testcontainer 1.15.0 via #363.

davecramer:
Ok so how does one do expectNext( some array ) ? The value returned is an array of strings with 3 elements. Alternatively how does one do expectNext(null) ?

mp911de:
`expectNext` does an `equals` check, so probably not ideal. Use `consumeNextWith` or `assertNext` as in:

```
            .assertNext(json -> {

                assertThat(json.data).contains("{ \"customer\": \"John Doe\"");

            })
```


Reactive Streams prohibits the usage of `null`. Instead of emitting null values, rather do not emit the element (or wrap it in `Optional`).

src/test/java/io/r2dbc/postgresql/ArrayAggIntegrationTests.java
davecramer: FYI, I copied `JsonIntegrationTests` which apparently don't subclass `AbstractIntegrationTests`

mp911de: I filed #364 to fix `JsonIntegrationTests`. I think, previously we had no config hook to force binary transfer.

davecramer:
So that works but it doesn't expose the bug. Thoughts ?

mp911de:
Adding the `NAME_ARRAY` seems about right. I'd go forward and merge this one. During the merge I'd also update the exception message to report the oid if a value cannot be decoded to aid debugging.

davecramer:
I'm not particularly happy about the name of the new  test. I'll change that and add the oid in the exception and merge it.

mp911de:
Great, thanks for letting me know. For issues and pull requests we use `[#???]` and `[resolves #???]` syntax to associate commit messages to tickets/pull requests.

davecramer:
What do you mean by cherry-picking ? You don't use the GUI to merge them ?

mp911de:
Exactly. I found that the easiest way is to `git cherry-pick <sha>` to bring commits into branches. This PR requires also backporting into `0.8.x` once it's merged.

PR #371, opened 2020-12-14
Fix readColumnFormats(...) to comply with postgresql decoding rules

### LINKED ISSUE #367: Decoding CopyOutResponse message does not comply with postgres documentation
#### Versions

- Driver: 0.8.X
- Database: docker:latest
- Java: 12
- OS: GNU/Linux

#### Current Behavior and Expected behavior/code

Documentation for `CopyOutResponse` at [postgres says:](https://www.postgresql.org/docs/current/protocol-message-formats.html)

> The format codes to be used for each column. Each must presently be zero (text) or one (binary). All must be zero if the overall copy format is textual.

And when I use `COPY some_table TO STDOUT;` I will receive byteBuff that has overall format of the text, but the code inside `AbstractCopyResponse` won't recognize this. In fact method `AbstractCopyResponse::readColumnFormats` in use case when the table has only two columns, for example, ID and some_text, it will receive count = 2 and in the switch statement it will ignore rest of the byteBuff which has two shortInt's with values of 0 and will go and set it to `Format.all()` which will make an assumption that it should be `Collections.unmodifiableSet(EnumSet.of(Format.FORMAT_BINARY, FORMAT_TEXT))`. Now maybe this is ok to work this way, but according to documentation and the decoding, the driver does not respect bytes sent by backend and instead of setting format to `Format.FORMAT_TEXT, FORMAT_TEXT` it set's the first one as binary.

#### Steps to reproduce

At forked repo I have added a unit test to reproduce the mentioned behavior:

- https://github.com/igor-susic/r2dbc-postgresql/blob/igor-susic/bug/decoding-copy-out-response/src/test/java/io/r2dbc/postgresql/message/backend/CopyOutResponseUnitTests.java#L59

And as I am trying to implement `COPY OUT` support, I need to know what is the way forward with this?

igor-susic:
Now AbstractCopyResponse.readColumnFormats(...) does not assume what to do when there are only one or two columns but instead reads buffer contents

[resolves #367]

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/message/backend/AbstractCopyResponse.java
+++ b/src/main/java/io/r2dbc/postgresql/message/backend/AbstractCopyResponse.java
@@ -84,18 +84,17 @@ abstract class AbstractCopyResponse implements BackendMessage {
         Assert.requireNonNull(in, "in must not be null");
 
         int count = in.readShort();
+        Set<Format> formatSet = EnumSet.noneOf(Format.class);
 
-        switch (count) {
-            case 0:
-                return EnumSet.noneOf(Format.class);
-            default:
-                Set<Format> formatSet = EnumSet.noneOf(Format.class);
-                for (int i = 0; i < count; i++) {
-                    formatSet.add(Format.valueOf(in.readShort()));
-                }
-
-                return formatSet;
+        if (0 == count) {
+            return formatSet;
         }
+
+        for (int i = 0; i < count; i++) {
+            formatSet.add(Format.valueOf(in.readShort()));
+        }
+
+        return formatSet;
     }
 
 }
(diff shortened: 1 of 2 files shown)


src/main/java/io/r2dbc/postgresql/message/backend/AbstractCopyResponse.java
```diff
@@ -88,11 +88,6 @@ public String toString() {
         switch (count) {
             case 0:
                 return EnumSet.noneOf(Format.class);
-            case 1:
```
mp911de, line 91: This could be even more simplified by removing `switch` entirely. When `count` is zero, then we don't even enter the loop.

mp911de (commented):
I left a comment to simplify the format decoding. Care to have a look?

src/main/java/io/r2dbc/postgresql/message/backend/AbstractCopyResponse.java
igor-susic: Sure, I agree it's better.

mp911de:

PR #373, opened 2020-12-18
Compat query mode

### LINKED ISSUE #341: Extended flow queries hang with pgpool 4.1
#### Versions

- Driver: 0.8.5.RELEASE
- Database: postgreSQL cluster with pgpool 4.1 (pgpool 4.0 works)
- Java: jdk8
- OS: win10

#### Current Behavior
I hava a PostgreSql cluster with Pgpool-II, other services connected to Pgpool using jdbc can run normally,  but service using 
r2dbc connect to pgpool, queries just hang forever.

mp911de:
`StatementCache` now no longer prepares statements (`Parse`) itself but rather reports whether statement preparation is required. That allows to streamline message handling to combine multiple messages into a composite one to reduce the number of sent TCP packets. `ExtendedFlowDelegate` encapsulates the extended flow.

We now support a compatibility mode for extended query flows. Enabling compatibility mode fetches all rows in auto-commit mode regardless the configured fetch size. Cursors are only used with explicit transactions and using `Sync` instead of `Flush` to work with newer pgpool versions.

Compatibility mode is disabled by default to retain the semantics of the previous driver version.

[fixes #341]


mp911de:
That's merged and backported now.

Issue #375, opened 2020-12-31
We need to add java.util.Currency support


DmitryErmolchik:
#### Is your feature request related to a problem? Please describe

I'm building a pet-project with java.util.Currency field and I was wondered r2dbc-postgresql doesn't support this basic data type.

#### Describe the solution you'd like

I propose to create a new codec **class CurrencyCodec extends AbstractCodec<Currency>** which will extract java.util.Currency.currencyCode and store it as a VARCHAR type.

#### Teachability, Documentation, Adoption, Migration Strategy

This is the new feature, no migration required.

mp911de:
We don't want to add type support in the core for every available Java core type, therefore, declining this feature request.

PR #376, opened 2020-12-31
Add support for java.util.Currency type

DmitryErmolchik:
#### Issue description

https://github.com/pgjdbc/r2dbc-postgresql/issues/375
 
#### New Public APIs

None


mp911de:
Not sure we want to have that. Such a codec should rather live in your own code.

DmitryErmolchik:
Currency is a Java core class. I'm sure this codec will be used in the pair with BigDecimalCodec.
Use-case I see is storing in the DB money amount as BigDecimal and currency as Currency. 
I think this will be good to have a codec to implement this experience out of the box.

mp911de:
You could argue the exact same way for every other Java class, Optional, ZoneId, Package, Cipher, Mac, and so on that can be obtained using a String. 

In that regard, we rather follow pgjdbc. If they support additional types that we don't, then it's fine to add an implementation. We set our to focus on core types which are pretty well supported and ans extension model that allows plugging in your own type support.

mp911de:
Closing with out merge as per comment above.

PR #398, opened 2021-04-23
Add Enum array codec

### LINKED ISSUE #391: Add enum array support
#### Versions

- Driver: 0.8.7.RELEASE

#### Current Behavior

According to the documentation, Postgres Enum Types are covered, by using EnumCodec.
Also, arrays are supported, like text[] type for example.

But I tried to have an array of enum in my model, and I am getting this error if I map to EnumSet<MyEnum> :

```java
Caused by: java.lang.IllegalArgumentException: Cannot decode value of type java.lang.Object with OID 16834
```

#### Table schema

<details>
<summary>Input Code</summary>

```sql
my_enum_field my_enum_type[] NOT NULL;
```

</details>

#### Steps to reproduce

<details>

in model class:

```java
private EnumSet<MyEnumType> myEnumField; // same with List<MyEnumType>
```
</details>

AbstractCoderX:
[resolves #391 ]

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/AbstractArrayCodec.java
@@ -52,20 +52,20 @@ abstract class AbstractArrayCodec<T> extends AbstractCodec<Object[]> {
 
     private final Class<T> componentType;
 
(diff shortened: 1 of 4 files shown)


mp911de:
Superseded by #399

PR #408, opened 2021-05-13
[Issue #407] Fix for HStore special characters

### LINKED ISSUE #407: Special characters inside HStore columns are not read correctly
#### Versions

- Driver: 0.8.8
- Database: 13.0
- Java: 15.0.1
- OS: Windows

#### Current Behavior

If an HStore column contains data with special characters like `"a"=>"Zoë", "é"=>"Foo"`, the special characters are not read correctly. It looks like some bytes are lost right after decode because the result of the example above will be `{"a"="Zoￃﾫ", "ￃﾩ"="Foo"}`.
 
#### Table schema

<details>
<summary>Input Code</summary>

```sql
CREATE EXTENSION IF NOT EXISTS hstore;

CREATE TABLE hstoretest (
  id SERIAL NOT NULL PRIMARY KEY,
  data HSTORE NOT NULL
);
```

</details>

#### Steps to reproduce

To reproduce the issue, create a table with an HStore column, create a row, and add some special characters to the HStore column (like `"a"=>"Zoë", "é"=>"Foo"`). Fetch the data from Java code and check the result.

#### Expected behavior/code

Special characters should be decoded correctly. This means a column containing `"a"=>"Zoë", "é"=>"Foo"` should decode to a Map containing `{"a"="Zoë", "é"="Foo"}`.

#### Possible Solution

Avoid casting from `byte` to `char` while parsing the data. Special characters are usually 2 bytes long and a cast can effectively cut off information from the character.

JoseLion:
Make sure that:

- [x] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes. (_I tried my best, but I use VSCode and I don't think the Intellij style file works there 😅_)

 

#### Issue description

Resolves #407

As mentioned in the ticket above, to parse the strings in the HStore data value, the code is reading the data byte by byte. However, to get a string out of the bytes, each byte is cast from `byte` to `char`, but characters can be 16-bits long rather than 8-bits.

To avoid the cast, I accumulated the parsed result in a `ByteBuffer` rather than the `StringBuilder` so we don't lose bytes in the parsing process. Then the bytes from `ByteBuffer` are used to create a UTF-8 encoded string.

I'm not sure if there's a better approach for this, but accumulating the bytes themselves is the only way I could think to preserve the integrity of the parsed byte. 🙂 

I added a couple of test cases with special characters to back the changes. Additionally, I cleaned up a few things I came across while working on this (boy scout principle 😅), which I'm fine reverting if necessary.

As always, I'm happy to contribute to this awesome project. 🙂 

Cheers!

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/HStoreCodec.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/HStoreCodec.java
@@ -26,7 +26,7 @@ import reactor.core.publisher.Mono;
 
 import javax.annotation.Nullable;
 
-import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -163,8 +163,8 @@ final class HStoreCodec implements Codec<Map> {
 
     @Nullable
     private static String readString(ByteBuf buffer) {
-        ByteBuffer accBuffer = ByteBuffer.allocate(buffer.capacity());
-        int position = buffer.forEachByte(new IndexOfProcessor((byte) '"'));
+        final ByteBuf accBuffer = buffer.alloc().buffer();
+        final int position = buffer.forEachByte(new IndexOfProcessor((byte) '"'));
 
         if (position > buffer.writerIndex()) {
             return null;
@@ -175,21 +175,19 @@ final class HStoreCodec implements Codec<Map> {
         }
 
         while (buffer.isReadable()) {
-            byte b = buffer.readByte();
-            char c = (char) b;
+            final byte current = buffer.readByte();
             
-            if (c == '"') {
+            if (current == '"') {
                 break;
             }
-            
-            if (c == '\\') {
-                b = buffer.readByte();
-            }
-            
-            accBuffer.put(b);
+
+            accBuffer.writeByte(current == '\\' ? buffer.readByte() : current);
         }
 
-        return new String(accBuffer.array(), 0, accBuffer.position(), UTF_8);
+        final String result = accBuffer.toString(StandardCharsets.UTF_8);
+
+        accBuffer.release();
+        return result;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/io/r2dbc/postgresql/codec/HStoreCodec.java
```diff
@@ -163,9 +162,10 @@ private static byte peekByte(ByteBuf buffer) {
     }
 
     @Nullable
-    private static String readString(StringBuilder mutableBuffer, ByteBuf buffer) {
-        mutableBuffer.setLength(0);
+    private static String readString(ByteBuf buffer) {
+        ByteBuffer accBuffer = ByteBuffer.allocate(buffer.capacity());
```
mp911de, line 166: This isn't necessary as a `ByteBuf` can be converted in to `String` by using `byteBuf.toString(StandardCharsets.UTF_8)`. We should not allocate arbitrary buffers but rather use `buffer.alloc().buffer(capacity)` to reuse the allocator and potentially pooled buffers. Make sure to `release` it after usage.

mp911de (changes requested):
See the comment I left regarding `ByteBuffer`

src/main/java/io/r2dbc/postgresql/codec/HStoreCodec.java
JoseLion: Awesome, thank you @mp911de! To be honest, I wasn't sure how to allocate a `ByteBuf`. This also helped to make the code a little simpler! I also added a few `final` in the local variables to avoid state mutation, let me know if that's ok 🙂 

I still have a doubt about the allocation, what would be the difference if we allocate the `accBuffer` with no initial capacity?
```java
ByteBuf accBuffer = buffer.alloc().buffer();
```

Tests still pass, so I assume the buffer will start with no initial capacity and grow as required, right? Would that be better or worse for performance? 🤔

mp911de:

PR #410, opened 2021-05-21
Introduce codec mapping caches

### LINKED ISSUE #409: Performance improvement on codec mapping
## Feature Request

While using this driver for fetching many rows from our table I noticed it was taking quite some time to build the `Row` objects. I hooked a profiler to the code and noticed it was spending quite some time for each row on the `codec.canDecode` method.
I wrote this little perf test to validate this finding:

```java
  @Test
  public void perfTestDecode() {
    DefaultCodecs codecs = new DefaultCodecs(TEST);
    long t = System.currentTimeMillis();
    for (int i = 0; i < 100000; i++) {
      for (int c = 0; c < 20; c++) {
        assertThat(
            codecs.decode(
                TEST.buffer(4).writeInt(100), INT4.getObjectId(), FORMAT_BINARY, Integer.class))
            .isEqualTo(100);
        assertThat(
            codecs.decode(
                ByteBufUtils.encode(TEST, "100"), INT2.getObjectId(), FORMAT_TEXT, Short.class))
            .isEqualTo((short) 100);
        assertThat(
            codecs.decode(
                ByteBufUtils.encode(TEST, "test"),
                VARCHAR.getObjectId(),
                FORMAT_TEXT,
                String.class))
            .isEqualTo("test");
        assertThat(
            codecs.decode(
                ByteBufUtils.encode(TEST, "2018-11-04 15:35:00.847108"),
                TIMESTAMP.getObjectId(),
                FORMAT_TEXT,
                LocalDateTime.class))
            .isInstanceOf(LocalDateTime.class);
        assertThat(codecs.decode(ByteBufUtils.encode(TEST, "{100,200}"), INT2_ARRAY.getObjectId(), FORMAT_TEXT, Object.class)).isEqualTo(new Short[]{100, 200});
        assertThat(codecs.decode(ByteBufUtils.encode(TEST, "{100,200}"), INT4_ARRAY.getObjectId(), FORMAT_TEXT, Object.class)).isEqualTo(new Integer[]{100, 200});
      }
    }
    System.out.println("Run in " + (System.currentTimeMillis() - t) + "ms");
  }
```
Running this on my machine results in:
```
Run in 13231ms
```
> NOTE That running the same example from 0.8.7.RELEASE results in 
> ```
> Run in 10389ms
> ```
> So it seems like the decoding performance went down between this release and the current master (at commit 9c773c8824b813e1ef65a26f18eadf97930a2933)

I would rather use this driver and webflux than going the web mvc route and jdbc.

I think the class `DefaultCodecs` would benefits from some simple caching mechanism. 
I did implement a simple caching (PR to come) and the results of the above test comes to:

```
Run in 7294ms
```

I haven't found alternatives

NA

frdeboffles:
Fixes #409

Introduce codec mapping caches to improve encoding and decoding performances.

* Switched the codec list to a thread safe variant to avoid the synchronized blocks. Even though `CopyOnWriteArrayList` is not super performant it should work fine in this context where the list should not be frequently updated.
* Switched `mockito-core` to `mockito-junit-jupiter` for Junit 5 support

Make sure that:

#### Issue description

See gh-409

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,8 +41,8 @@
         <junit.version>5.7.0</junit.version>
         <jmh.version>1.32</jmh.version>
         <mbr.version>0.2.0.RELEASE</mbr.version>
-        <logback.version>1.2.3</logback.version>
-        <mockito.version>3.11.2</mockito.version>
+        <logback.version>1.2.5</logback.version>
+        <mockito.version>3.12.4</mockito.version>
         <netty.version>4.1.66.Final</netty.version>
         <postgresql.version>42.2.23</postgresql.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -265,7 +265,7 @@
                 <version>3.8.1</version>
                 <configuration>
                     <compilerArgs>
-<!--                        <arg>-Werror</arg>-->
+                        <arg>-Werror</arg>
                         <arg>-Xlint:all</arg>
                         <arg>-Xlint:-deprecation</arg>
                         <arg>-Xlint:-options</arg>
--- a/pom.xml
+++ b/pom.xml
@@ -210,7 +210,7 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
+            <artifactId>mockito-junit-jupiter</artifactId>
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
@@ -265,7 +265,7 @@
                 <version>3.8.1</version>
                 <configuration>
                     <compilerArgs>
-                        <arg>-Werror</arg>
+<!--                        <arg>-Werror</arg>-->
                         <arg>-Xlint:all</arg>
                         <arg>-Xlint:-deprecation</arg>
                         <arg>-Xlint:-options</arg>
--- a/src/jmh/java/io/r2dbc/postgresql/CodecRegistryBenchmarks.java
+++ b/src/jmh/java/io/r2dbc/postgresql/CodecRegistryBenchmarks.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.r2dbc.postgresql;
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.UnpooledByteBufAllocator;
+import io.r2dbc.postgresql.codec.CodecFinderCacheImpl;
+import io.r2dbc.postgresql.codec.CodecFinderDefaultImpl;
+import io.r2dbc.postgresql.codec.Codecs;
+import io.r2dbc.postgresql.codec.DefaultCodecs;
+import io.r2dbc.postgresql.util.ByteBufUtils;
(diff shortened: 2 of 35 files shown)


mp911de:
Thanks a lot for submitting a pull request. Introducing caching comes with quite some complexity. The on-demand cache registration of codecs resonates with my cautionary sense that the state of the codec registry changes over time (as the application keeps running) which can lead to some point to driver behavior that cannot be properly reproduced because of so many moving parts.

I'd like much more to explore a path where we pre-populate the caches to keep a stable state on a per-connection level. With doing so, potential bugs can be caught early on as each connection enters a stable state after establishing the connection.

I think to make that work, we would need to extend our codecs with some sort of interface that exposes which data types, classes and formats are supported (`CodecMetadata.getFormats()/.getDataTypes()/…`).

frdeboffles:
Sorry for the late reply. I will have a look at pre-filling the caches. This does imply that we don't support a codec that can encode sub-classes of the associated type.

frdeboffles:
@mp911de I followed your recommendations and tried to implement a pre-filled cache. 
This time the changes I made are more invasive than before...
One issue is that we can't predict all the possible combinations given the support for nested arrays for example. I tried to cover 1 dimensional arrays in this implementation. 
Please let me know if this is what you had in mind?

mp911de:
Thanks a lot. I'm not sure about the codec finder API yet, but having each codec exposing details which data types it is able to process makes a lot of sense.

Are you able to come up with some benchmarks to see whether this optimization makes sense? We have some integrative benchmarks in the `src/jmh` directory. For the codecs, we could have benchmarks that do not require database interaction but rather work with the codec API directly.

frdeboffles:
I'll add benchmarks that compare the encode and decode using both the cache and default implementations of the codec finder.
About the `CodecFinder` API, my thinking was to offer the users the ability to use or not the cache based finder. I guess the cache based finder may come slower in some particular case where an application is constantly requesting a new connection.

frdeboffles:
I added some benchmarks as follow in the last commit:
```
Benchmark                                                (iterations)   Mode  Cnt       Score      Error  Units
CodecRegistryBenchmarks.decodeWithCacheDisabledRegistry            10  thrpt    5   15525.438 ± 3219.915  ops/s
CodecRegistryBenchmarks.decodeWithCacheDisabledRegistry           100  thrpt    5    1438.332 ±   53.718  ops/s
CodecRegistryBenchmarks.decodeWithCacheDisabledRegistry          1000  thrpt    5     134.925 ±    2.589  ops/s
CodecRegistryBenchmarks.decodeWithCacheEnabledRegistry             10  thrpt    5   28435.893 ± 1787.007  ops/s
CodecRegistryBenchmarks.decodeWithCacheEnabledRegistry            100  thrpt    5    2859.265 ±   82.144  ops/s
CodecRegistryBenchmarks.decodeWithCacheEnabledRegistry           1000  thrpt    5     282.089 ±    9.164  ops/s
CodecRegistryBenchmarks.encodeWithCacheDisabledRegistry            10  thrpt    5   34916.352 ± 1826.371  ops/s
CodecRegistryBenchmarks.encodeWithCacheDisabledRegistry           100  thrpt    5    3505.425 ±  184.198  ops/s
CodecRegistryBenchmarks.encodeWithCacheDisabledRegistry          1000  thrpt    5     341.682 ±   29.398  ops/s
CodecRegistryBenchmarks.encodeWithCacheEnabledRegistry             10  thrpt    5  160630.081 ± 8314.723  ops/s
CodecRegistryBenchmarks.encodeWithCacheEnabledRegistry            100  thrpt    5   15919.121 ±  578.995  ops/s
CodecRegistryBenchmarks.encodeWithCacheEnabledRegistry           1000  thrpt    5    1572.680 ±   58.117  ops/s
```

mp911de:
Wow, that's amazing, 2x up to 10x improvement in throughput.

frdeboffles:
I originally tested this on the `0.8.x` branch but this was the original commit of this PR. 
I can try porting this in another PR.

mp911de:
Thanks a lot. I'll come back to this next week.

mp911de:

frdeboffles:
Thank you for accepting it. I like your version better ;)

mp911de:
Thanks for having a look at the polishing commits, I'll incorporate your feedback.

Issue #420, opened 2021-07-01
PostgresObjectId#isValid uses exception as control mechanism resulting in significant performance penalty


amoffet:
#### Versions

- Driver: 0.9.0.M1
- Database: Postgres 12
- Java: 1.8
- OS: OSX

#### Current Behavior

When using a custom type such as ENUM for a column, it will be assigned an object id that is not in the usual range.  The code in PostgresObjectId#isValid looks like this:

```
        if (objectId >= 0 && objectId < OID_CACHE_SIZE) {
            PostgresqlObjectId oid = CACHE[objectId];
            return oid != null;
        }

        try {
            valueOf(objectId);
            return true;
        } catch (Exception e) {
            return false;
        }
```

As a result, any row that has a custom type (such as enum) will now go through the exception path (to return false), which is expensive.  In my particular case, it results in 88,877 exceptions in 2 minutes.

It seems like #isValid should not depend on the exception throwing behavior of #valueOf and should return true or false directly.

<details>
<summary>Stack trace</summary>

```
// your stack trace here
```

</details>
 
#### Table schema

<details>
<summary>Input Code</summary>

```sql
-- your SQL here;
```

</details>

#### Steps to reproduce

<details>
<summary>Input Code</summary>

```java
// your code here;
```

</details>

amoffet:
Perhaps something as easy as:

```
    public static boolean isValid(int objectId) {

        if (objectId >= 0 && objectId < OID_CACHE_SIZE) {
            PostgresqlObjectId oid = CACHE[objectId];
            return oid != null;
        }

        for (PostgresqlObjectId type : values()) {
            if (type.objectId == objectId) {
                return true;
            }
        }
        return false;
    }
```
which is essentially the same code...

mp911de:
Thanks for reporting the issue. Calling values() copies the internal value array so the check isn't ideal. Maybe we can increase the cache size to cover the full range and get away with a bit increased memory size in return for avoiding exceptions and allocations on the hot code path.

Squiry:
Why do we even bother iterating values with known ids? No value can be there that is greater than `OID_CACHE_SIZE`. Block
```java
try {
    valueOf(objectId);
    return true;
} catch (Exception e) {
    return false;
}
```
always returns false now, doesn't it?

mp911de:
[...]

PR #421, opened 2021-07-01
Replace use of exceptions as control mechanism

amoffet:
We now look through the list of PostgresqlObjectids to determine if a given objectId is valid - Instead of using the exception-throwing behavior of #valueOf

See [#420](https://github.com/pgjdbc/r2dbc-postgresql/issues/420)


amoffet:
Closing - Calling values() on the Enum causes a full copy.

PR #438, opened 2021-08-24
 Avoid cast to X509Certificate array during certificate verification

Deniz97:
Make sure that:

- [ ] You have created a feature request first to discuss your contribution intent. Please reference the feature request ticket number in the pull request. (No, trivial change)
- [ ] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.

 

#### Issue description

Hey, so the array cast in the old version fails for us, when we enable `ssl` trough application.yml. I don't know if it ever worked for us before, but here is the related StackOverflow discussion that led us to this fix: https://stackoverflow.com/questions/37248975/error-ljava-lang-object-cannot-be-cast-to-ljava-security-cert-x509certificate
 
 This version do indeed work for us, the array cast is the problem, casting the elements one by one solves it. I did an oldschool for loop, the code can be improved.
 
 
#### New Public APIs

None

#### Additional context

None

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/client/DefaultHostnameVerifier.java
+++ b/src/main/java/io/r2dbc/postgresql/client/DefaultHostnameVerifier.java
@@ -28,7 +28,6 @@ import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
 import javax.security.auth.x500.X500Principal;
 import java.net.IDN;
-import java.security.cert.Certificate;
 import java.security.cert.CertificateParsingException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
--- a/src/main/java/io/r2dbc/postgresql/client/DefaultHostnameVerifier.java
+++ b/src/main/java/io/r2dbc/postgresql/client/DefaultHostnameVerifier.java
@@ -99,22 +99,6 @@ public enum DefaultHostnameVerifier implements HostnameVerifier {
 
     @Override
     public boolean verify(String hostname, SSLSession session) {
-        X509Certificate[] peerCerts;
-        try {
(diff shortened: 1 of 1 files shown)


frankgh:
@Squiry are you able to provide a repro for this issue?

Deniz97:
Hey guys,

So here is a little example, i could not did the same with Certificate classes because they are abstract.

jdoodle.com/ia/gka (Code below)

I can also provide a full repo with micronaut and testcontainers if asked, no problem, just a little busy so it would be ready to weekend, but array casting seems to be really problematic in java.

```
import java.util.Arrays;

public class MyClass {
    public static void main(String args[]) {
      Integer[] ints = {11, 22, 33};
      // Number[] numbers = (Number[]) ints; // ERROR
      Number[] numbers = Arrays.copyOf(ints, ints.length, Number[].class);
      System.out.println("result: " + numbers.toString());
      
      // Integer[] ints2 = (Integer[]) numbers; // ALSO ERROR
      // Integer[] ints2 = Arrays.copyOf(numbers, numbers.length, Integer[].class); // ALSO ERROR
      
      // So even with Arrays.copyOf, one can upcast but cannot downcast. Certificate -> X509Certificate is a downcast.
      
    }
}
```

Squiry:
I can't remember if I ever met any `Certificate[]` in a wild but yeah, there can be a `ClassCastException` in that case. No need to copy all array and cast every element though: you can cast just first element here `X509Certificate serverCert = peerCerts[0];`(line 134).

Deniz97:
@Squiry did that, thanks.

mp911de:

Issue #483, opened 2022-01-24
Support for postgis geometry parent type


javieronsurbe:
## Feature Request
PostGIS have support to define a generic geometry type for fields that can contains points, polygons, etc. and this kind of field is supported in Postgres JDBC version using an [abstract class](https://github.com/postgis/postgis-java/blob/main/postgis-jdbc-geometry/src/main/java/net/postgis/jdbc/geometry/Geometry.java) that specific ones can inherit.

Can we have some similar behavior in r2dbc one? I.e.-  Declaring a field in a POJO of type Geometry (abstract) and when it is recovered from database from a table field of type geometry it is mapped to a specific type (Point, Polygon, etc.) 

As I see specific ones support are added (0.8.5 version), maybe it's not very difficult to add this enhancement to driver

Thanks!

mp911de:
R2DBC's extension model is based on codecs so plugin authors can bring their own type hierarchy. Looking though the code of Postgis JDBC, I haven't found how Postgis registers itself with the JDBC driver. I think you could reuse the same types and provide a Postgis codec to natively read into Geometry types and write such.

Take a look at the `HSTORE` codec that is being registered conditionally if `HSTORE` is present: https://github.com/pgjdbc/r2dbc-postgresql/blob/main/src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java#L32

javieronsurbe:
Thank you very much for the hint Mark. I will try it.

PR #491, opened 2022-02-06
Add PostgisGeometryCodec

earlbread:
Make sure that:

#### Issue description

This supports Postgis Geometry type. #483

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
@@ -33,9 +33,24 @@ public class BuiltinDynamicCodecs implements CodecRegistrar {
 
     private static final Object EMPTY = new Object();
 
-    enum BuiltinCodec {
+    private interface CodecSupport {
+        default boolean isSupported() {
+            return true;
+        }
+    }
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
```diff
@@ -48,6 +49,8 @@
             switch (this) {
                 case HSTORE:
                     return new HStoreCodec(byteBufAllocator, oid);
+                case POSTGIS_GEOMETRY:
```
mp911de, line 52: The code should include a guard to verify that the required classes for `jts-core` are on the class path.

earlbread: Added `isSupported` method and changed to register codec only when it returns `true`.

earlbread:
Since `WkbReader` is not thread-safe, I changed `decode` to create a `WkbReader` instance every time.

mp911de:
Codecs aren't really used concurrently, they are rather used sequentially and each connection maintains its own codec instances. In any case, codecs might be used by different threads and that can introduce visibility issues so it's perfectly fine to maintain individual instances.

mp911de:

PR #495, opened 2022-02-16
Fix postgres OID type decoding

k-tokarev:
#### Issue description

If returned OID is big enough (2314556683 for example) codec throws NumberFormatException. 
 
#### New Public APIs

No new public APIs added

#### Additional context

According to documentation (https://www.postgresql.org/docs/12/datatype-oid.html) OID type is "an unsigned four-byte integer", so it is incorrect to decode it as java Integer (which is signed an so has lower max value). 

This pull request decodes OIDs as Long to fix the problem


mp911de:

Issue #183, opened 2019-10-18
Add `COPY FROM` support


Globegitter:
I have not been able to find anything in the documentation or through the github code-search, but does this driver support the `COPY FROM` command? Here is the official documentation of the command that I am talking about: https://www.postgresql.org/docs/current/sql-copy.html

mp911de:
What exactly should the driver do? It's just an SQL command, right? You might want to consider pgjdbc/pgjdbc#1299 for the `stdin` case. There's no `CopyManager` API that allows copy in/copy out operations. We're happy to review a pull request to make these things happen.

Globegitter:
Thanks for the pointer - I am not sure I can really answer your question, if it should just be working then it might just be worth adding a note about in that in the documentation or adding a test for it. Part of the problem is I was comparing a few different drivers and so far the ones that I have tried have not supported it. After I found this one I could not find anything looking at the website / searching through the repository here. So based on my experience so far I assumed it was not supported.

mp911de:
You'd help us by outlining your use-case and what you're trying to achieve to get a better impression of what's needed to make `COPY FROM` work, ideally with a bit of SQL.

igor-susic:
@mp911de If no one is working on this one, I'm more than willing to submit PR.

mp911de:
Thanks for your support. Feel free to come up with a design proposal so we can discuss the overall direction before submitting a pull request.

igor-susic:
Sure, no problem, just give me a little to get to know the codebase better and figure out the proposal :)

ArjanSchouten:
@igor-susic is there any progress on this issue?

mp911de:
As per https://github.com/pgjdbc/r2dbc-postgresql/issues/183#issuecomment-716173895, we're happy to review proposals if you have a use case for a reactive `COPY FROM`.

ArjanSchouten:
The use case is a lot of data from a REST interface that has to be inserted into a Greenplum database. Greenplum and lots of insert statements is slow and `COPY FROM` is the best performant way to insert a lot of data.

I don't know where to start to implement this and don't have much time to figure out how the JDBC driver is currently doing it.

mp911de:
It sounds as if the API would need to consume a `Publisher<ByteBuffer>` as a data source that is then piped into the database.

ArjanSchouten:
Ok so I currently use this:
```
[...]

PR #500, opened 2022-03-18
Add `COPY FROM` support

ArjanSchouten:
Add copy in support for r2dbc postgresql driver. Postgresql supports COPY statements to insert data fast: https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-COPY

The implementation is inspired by [CopyManager](https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/copy/CopyManager.java) and [QueryExecutor](https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/core/v3/QueryExecutorImpl.java) from https://github.com/pgjdbc/pgjdbc

#### Issue description

This PR contains the code for issue https://github.com/pgjdbc/r2dbc-postgresql/issues/183.
 
#### New Public APIs

```
    /**
     * Copy bulk data from client into a PostgreSQL table very fast.
     *
     * @param sql the COPY sql statement
     * @param stdin the ByteBuffer publisher
     * @return a {@link Mono} with the amount of rows inserted
     */
    Mono<Long> copyIn(String sql, Publisher<ByteBuffer> stdin);
```

#### Additional context

Note: we added a benchmark test. It turned out that reading the csv files causes much of the differences in our testing.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlCopyIn.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlCopyIn.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.r2dbc.postgresql;
+
+import io.netty.buffer.Unpooled;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
+import io.r2dbc.postgresql.client.Client;
+import io.r2dbc.postgresql.message.backend.BackendMessage;
+import io.r2dbc.postgresql.message.backend.CommandComplete;
+import io.r2dbc.postgresql.message.backend.CopyInResponse;
+import io.r2dbc.postgresql.message.backend.ReadyForQuery;
+import io.r2dbc.postgresql.message.frontend.CopyData;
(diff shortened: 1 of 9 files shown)


src/main/java/io/r2dbc/postgresql/api/PostgresqlConnection.java
```diff
@@ -170,4 +172,13 @@
     @Override
     Mono<Boolean> validate(ValidationDepth depth);
 
+    /**
+     * Copy bulk data from client into a PostgreSQL table very fast.
```
mp911de, line 176: Nit: We should mention `COPY … FROM STDIN` here as well.

mp911de (commented):
Thanks a lot for your contribution. The API looks pretty decent. I added a few comments. Care to have a look?

src/main/java/io/r2dbc/postgresql/PostgresqlCopyIn.java
```diff
@@ -0,0 +1,101 @@
...
+            .map(buffer -> new CopyData(Unpooled.wrappedBuffer(buffer)))
+            .as(messages ->  copyIn(sql, messages));
+    }
+
+    private Mono<Long> copyIn(String sql, Flux<CopyData> copyDataMessages) {
+        Client client = context.getClient();
+
+        Flux<BackendMessage> backendMessages = copyDataMessages
+            .doOnNext(client::send)
+            .doOnError(e -> !(e instanceof IllegalArgumentException), (e) -> sendCopyFail(e.getMessage()))
```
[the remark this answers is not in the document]
ArjanSchouten: Removed, it was meant for the IllegalArgumentException thrown in startCopy but I moved the doOnError onto the backendMessages only. Thanks

src/main/java/io/r2dbc/postgresql/api/PostgresqlConnection.java
ArjanSchouten: I did copy it from pgjdbc's CopyManager 😄

mp911de:
Not yet, but I'd like to include this API for the 1.0 release. @harmvanderwal did you get a chance to build the PR locally so you can verify the API is suitable for your use case?

harmvanderwal:
I did and it works perfect for my specific use case.

mp911de:

PR #504, opened 2022-04-05
Add `sslMode` option lowercase alias

### LINKED ISSUE #486: Consider lowercase `sslmode` for easier JDBC URL adaption
## Bug Report

Migrating from jdbc to r2dbc is confusing due to the jdbc parameter of `sslmode` and r2dbc requiring it to be cased differently as `sslMode` -- Ideally they should be the same, but at a minimum, i'd suggest a warning during URL parsing to indicate that the case differs.

#### Current Behavior

parameter `sslmode` does not work.

parameter `sslMode` works.

ssttaarr:
For easier JDBC URL adaption

[resolves #486]


mp911de:

PR #510, opened 2022-05-05
Add support to parse timestamps with `+hh:mm:ss` offset

### LINKED ISSUE #509: Issue parsing timestamp with timezone having offset of format `+hh:mm:ss`

tested on `postgres:10-alpine` docker image

I have an issue when trying to parse `timestamp with timezone` in format of `0001-01-01 08:20:53+02:30:17` to `LocalDateTime`/`OffsetDateTime`/`ZonedDateTime`. Here is an example test code on groovy:

```groovy
def temp() {
        Function<Row, OffsetDateTime> fn = note -> note.get("datetime", OffsetDateTime)

        expect:
        def datetime = databaseClient.sql("SELECT '0001-01-01 08:20:53+02:30:17'::timestamptz as datetime")
            .map(fn).first()
            .block()
    }
```

which giving following stacktrace:
```log
Text '0001-01-01 08:20:53+02:30:17' could not be parsed, unparsed text found at index 25
java.time.format.DateTimeParseException: Text '0001-01-01 08:20:53+02:30:17' could not be parsed, unparsed text found at index 25
	at java.base/java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:2055)
	at java.base/java.time.format.DateTimeFormatter.parse(DateTimeFormatter.java:1954)
	at io.r2dbc.postgresql.codec.AbstractTemporalCodec.decodeTemporal(AbstractTemporalCodec.java:135)
	at io.r2dbc.postgresql.codec.AbstractTemporalCodec.decodeTemporal(AbstractTemporalCodec.java:95)
```

it seems that for year which is lower then 1920 postgres does not convert timestamp with timezone to format of `+hh:00:00` and leaves it in unformatted way (e. g. `+02:30:17`).
This could be fixed by adding following change to [PostgresqlDateTimeFormatter](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/src/main/java/io/r2dbc/postgresql/codec/PostgresqlDateTimeFormatter.java)

```diff
-            .appendOffset("+HH:mm", "+00:00")
+            .appendOffset("+HH:mm:ss", "+00:00:00")
```

Conversion of offset could be fixed in postgresql which will take more time, I suppose

Nothing to document

khlopkov:
Make sure that:

#### Issue description

Previously parsing timestamp with timezone failed if offset had format `+hh:mm:ss`. This PR fixes it
Fixes #509  

#### New Public APIs

There are no new APIs.

#### Additional context

Issue was encountered on `postgresql:10-alpine` docker image

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/PostgresqlDateTimeFormatter.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/PostgresqlDateTimeFormatter.java
@@ -48,7 +48,7 @@ class PostgresqlDateTimeFormatter {
             .appendFraction(NANO_OF_SECOND, 0, 9, true)
             .optionalEnd()
             .optionalStart()
-            .appendOffset("+HH:mm", "+00:00")
+            .appendOffset("+HH:mm:ss", "+00:00:00")
             .optionalEnd()
             .toFormatter();
 
--- a/src/test/java/io/r2dbc/postgresql/codec/InstantCodecUnitTests.java
+++ b/src/test/java/io/r2dbc/postgresql/codec/InstantCodecUnitTests.java
@@ -66,6 +66,8 @@ final class InstantCodecUnitTests {
         ZonedDateTime zonedDateTime = ZonedDateTime.parse("2018-11-05T00:20:25.039883+09:00[Asia/Tokyo]");
         Instant instant = zonedDateTime.toInstant();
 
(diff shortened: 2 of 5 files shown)


mp911de:
The changes above break the build because `+HH:mm:ss` is not a supported pattern. `+HH:MM:ss` (uppercase M) works but that breaks other test. The contribution needs a bit more work, currently looking into it.

PR #516, opened 2022-05-24
Update Netty to 4.1.77.Final with a matching Reactor version

### LINKED ISSUE #514: Upgrade to netty 4.1.77.Final to address epoll_wait() not implemented on Linux
#### Versions

- Driver: 1.0.0.BUILD-SNAPSHOT
- Database: PostgreSQL 14
- Java: OpenJDK 11.0.15
- OS: Manjaro Linux (5.4.192-1-MANJARO)

#### Current Behavior

On my Linux machine with the current version of `r2dbc-postgresql` on master (1.0.0.BUILD-SNAPSHOT built locally) my database calls fail with the following error:
```
io.netty.channel.epoll.EpollEventLoop handleLoopException
WARNING: Unexpected exception in the selector loop.
io.netty.channel.unix.Errors$NativeIoException: epoll_wait(..) failed: Function not implemented
``` 
#### Steps to reproduce

I suppose any query will have this problem on Linux, but I used this in particular in Scala 3
```scala
val result = factory.create.flux
    .flatMap(_.createStatement("DELETE FROM items").execute)
    .flatMap(_.getRowsUpdated)
    .collectList.block
```

And here's the link to a repo with the example https://github.com/jczuchnowski/r2dbc-example

#### Possible Solution

This problem was probably solved here https://github.com/netty/netty/issues/12343 with version of Netty 4.1.77.Final but `r2dbc-postgresql` is transitively using 4.1.76.Final.

jczuchnowski:
Resolves #514 

Updates Netty to 4.1.77.Final along with reactor-bom to 2020.0.19. It seems like 2020.0.19 is newer than 2022.0.0-M2 and contains compatible Netty dependencies.


mp911de:
Did you test whether the newer netty version works against Reactor Netty 2022.0.0-M2? There should be no need to upgrade the Reactor version for now.

mp911de:
Care to update the PR to include just the netty version change and roll back the reactor change?

jczuchnowski:
Done.

mp911de:

PR #584, opened 2023-02-15
Bump Netty to 4.1.89.Final, CVE-2022-41915

patriknw:
* The CVE is probably not related to how Netty is used in r2dbc-postgresql but nice to use latest patch to avoid warnings from security scanning tools.


mp911de:
Thank you for your contribution. That's merged now.

PR #594, opened 2023-06-07
Adapt integration test to be Java 11 forward-compatible

fkomauli:
Reintroduced JDK > 8 incompatible integration test with java.time.Instant stored as timestamptz.
Instant.now() precision changed from ms to ns after JDK 8, so tests must deal with conversions to postgres μs


mp911de:

PR #606, opened 2023-07-16
Reduce time complexity for PostgresqlRow.getColumn(String name) method

gc-garcol:
Tuning PostgresqlRow.getColumn(String name) method

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlRow.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlRow.java
@@ -27,10 +27,14 @@ import io.r2dbc.spi.Row;
 import reactor.core.publisher.Mono;
 import reactor.util.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


gc-garcol:
missing check equalsIgnoreCase for field name

src/main/java/io/r2dbc/postgresql/PostgresqlRow.java
```diff
@@ -165,25 +172,12 @@ void release() {
...
 
     private List<String> toColumnNames() {
-        List<String> names = new ArrayList<>(this.fields.size());
-
-        for (RowDescription.Field field : this.fields) {
-            names.add(field.getName());
-        }
-
-        return names;
+        return fields.stream().map(RowDescription.Field::getName).collect(Collectors.toList());
```
mp911de, line 180: Generally speaking, the Stream API adds a lot of GC and memory pressure.

PR #608, opened 2023-08-27
Add codecs for `DayOfWeek`, `Month`, `MonthDay`, `Period`, `Year`, `YearMonth`

### LINKED ISSUE #591: Encode/Decode of other Java 8 DateTime types
I tried to use `Year` type(`INT` in the db column) in my projects(Spring Boot 3.0.4, R2dbc 1.0),  and got the following exceptions:

```bash
java.lang.IllegalArgumentException: Cannot encode parameter of type java.time.Year (2023)
```

I think all missing Java 8 DateTime types should be supported in Postgres R2dbc drivers.

* Year
* Month
* DayOfWeek
* DayOfMonth
* YearMonth
* etc.

steverigney:
Make sure that:

- [ ] You have read the [contribution guidelines](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/.github/CONTRIBUTING.adoc).
- [ ] You have created a feature request first to discuss your contribution intent. Please reference the feature request ticket number in the pull request.
- [ ] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.


src/main/java/io/r2dbc/postgresql/codec/DefaultCodecs.java
```diff
@@ -96,97 +96,6 @@ public DefaultCodecs(ByteBufAllocator byteBufAllocator, boolean preferAttachedBu
         this.codecLookup.afterCodecAdded();
     }
 
-    @SuppressWarnings({"unchecked", "rawtypes"})
```
mp911de, line 99: What's the reason for this change? It makes the pull request almost illegible because it mixes refactorings with actual changes.

mp911de (changes requested):
The actual code changes look quite decent. However, there's a log of refactoring changes. Please either roll these back, specifically `DefaultCodecs`, or move these into a separate commit for easier review.

src/main/java/io/r2dbc/postgresql/codec/DefaultCodecs.java
steverigney: This looks to have been introduced as part of accidental reformatting and is an oversight on my part, I have created a new commit that rolls back the formatting changes to Default Codec.

PR #632, opened 2024-01-29
Do not require typarray column when auto-registering extensions

chrischall:


src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
```diff
@@ -119,7 +119,7 @@ public Publisher<Void> register(PostgresqlConnection connection, ByteBufAllocato
     }
 
     private PostgresqlStatement createQuery(PostgresqlConnection connection) {
-        return connection.createStatement(String.format("SELECT oid, typname, typarray FROM pg_catalog.pg_type WHERE typname IN (%s)", getPlaceholders()));
+        return connection.createStatement(String.format("SELECT oid, typname, * FROM pg_catalog.pg_type WHERE typname IN (%s)", getPlaceholders()));
```
mp911de, line 122: Care to rewrite the SQL into `SELECT *` as that is a simpler variant?

PR #640, opened 2024-02-24
Add column name index mapping in `PostgresqlRow`

### LINKED ISSUE #636: Performance issue with PostgresqlRow.getColumn(String name) when select many columns
When selecting many fields, a lot of CPU is consumed because the fields are looped.

In one case I encountered, with a select of millions of records with about 700 fields, I forced the query to stop because it was consuming 100% CPU for several hours.

At this time, I checked Java's stack trace many times, and most of them were processing the following for statement.

https://github.com/pgjdbc/r2dbc-postgresql/blob/d047276aec03b9d691cb343789dc8d933e0dad7f/src/main/java/io/r2dbc/postgresql/PostgresqlRow.java#L168-L174

Would it be possible to prepare field name and order mappings in advance?

cty123:
[resolves #636]

Make sure that:

- [ x ] You have read the [contribution guidelines](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/.github/CONTRIBUTING.adoc).
- [ x ] You have created a feature request first to discuss your contribution intent. Please reference the feature request ticket number in the pull request.
- [ x ] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.
- [ x ] You submit test cases (unit or integration tests) that back your changes.

 

#### Issue description

Previously getting the fields of the row results requires looping the fields array list to find the field with the specified name, this can be very slow when there's a large amount of columns retrieved. Here I am trying to add a cache to map the column name to index such that we only need to loop the fields list once.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlRow.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlRow.java
@@ -24,6 +24,9 @@ import io.r2dbc.postgresql.message.backend.DataRow;
 import io.r2dbc.postgresql.message.backend.RowDescription;
 import io.r2dbc.postgresql.util.Assert;
 import io.r2dbc.spi.Row;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
 import reactor.core.publisher.Mono;
 import reactor.util.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


mp911de:

PR #641, opened 2024-03-03
Reuse column index mapping from `PostgresqlColumnMetadata`

### LINKED ISSUE #636: Performance issue with PostgresqlRow.getColumn(String name) when select many columns

cty123:
[resolves #636]

Make sure that:

- [ x ] You have read the [contribution guidelines](https://github.com/pgjdbc/r2dbc-postgresql/blob/main/.github/CONTRIBUTING.adoc).
- [ x ] You have created a feature request first to discuss your contribution intent. Please reference the feature request ticket number in the pull request.
- [ x ] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.
- [ x ] You submit test cases (unit or integration tests) that back your changes.

 

#### Issue description

Following up on the previous change #640 , the field name look up is still not optimal for usecases where the result contains a large chunk of data. Take [r2dbc example usecase](https://r2dbc.io/),

```java
Mono.from(connectionFactory.create())
  .flatMapMany(connection -> connection
    .createStatement("SELECT firstname FROM PERSON WHERE age > $1")
    .bind("$1", 42)
    .execute())
  .flatMap(result -> result
    .map((row, rowMetadata) -> row.get("firstname", String.class)))
  .doOnNext(System.out::println)
  .subscribe();
```

imagine the result set returns 1 million rows of data, because currently the row level column name index map is computed on the fly per row, the same mapping would need to be computed 1 million times, which is pretty much futile work, as we only need to compute the mapping once. Meanwhile, looking at the implementation of pgjdbc, the mapping is computed per result set.

Another small change I made is to avoid putting caches on case matches, and defaults the match to lower case. Since metadata object now owns the column name lookup cache, and each row might be processed concurrently, the cache can have concurrency problem when multiple threads want to put cache simultaneously. So I am now making it read only.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlRow.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlRow.java
@@ -50,13 +50,21 @@ final class PostgresqlRow implements io.r2dbc.postgresql.api.PostgresqlRow {
 
     private volatile boolean isReleased = false;
 
-    private Map<String, Integer> columnNameIndexCacheMap;
+    private final Map<String, Integer> columnNameIndexCacheMap;
 
     PostgresqlRow(ConnectionResources context, io.r2dbc.postgresql.api.PostgresqlRowMetadata metadata, List<RowDescription.Field> fields, ByteBuf[] data) {
         this.context = Assert.requireNonNull(context, "context must not be null");
         this.metadata = Assert.requireNonNull(metadata, "metadata must not be null");
         this.fields = Assert.requireNonNull(fields, "fields must not be null");
         this.data = Assert.requireNonNull(data, "data must not be null");
+
+        if (metadata instanceof PostgresqlRowMetadata) {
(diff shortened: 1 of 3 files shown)


mp911de:
Thanks for the suggestion. It makes totally sense to add this optimization.

Issue #631, opened 2024-01-29
Do not require `typarray` column when auto-registering extensions


chrischall:
See #612 and #621

mp911de:
Fixed via 24af09455e8d281c57705bcbc9189eddbfbb8ed1. Care to test against the latest 1.0.5 snapshot? (`1.0.5.BUILD-SNAPSHOT`)

PR #649, opened 2024-04-10
Only get oid and typarray when extension is supported.

chrischall:
Addresses #631.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
@@ -104,12 +104,12 @@ public class BuiltinDynamicCodecs implements CodecRegistrar {
         return statement.execute()
             .flatMap(it -> it.map((row, rowMetadata) -> {
 
-                    int oid = PostgresqlObjectId.toInt(row.get("oid", Long.class));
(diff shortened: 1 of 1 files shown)


mp911de:
Care to elaborate what this change achieves other than delaying the lookup?

chrischall:
> Care to elaborate what this change achieves other than delaying the lookup?

Apologies. Our application connects to Redshift which is compatible with PostgreSQL 8.0.2. Unfortunately, that version does not have `oid` or `typarray` columns in the `pg_catalog.pg_type` table. There was already a fix to set the type to `PostgresTypes.NO_SUCH_TYPE` in the event the `typarray` column is missing. However, there was no similar fix for the missing `oid` column.

This fix handles that by only attempting to get the `oid` column is the `BuiltinCodec` is supported. For the `POSTGIS_GEOMETRY` codec, this relies on another library being on the classpath. So, for our application, we do not have that library on our classpath so it will not try to register that codec and will not try to get the `oid` column.

mp911de:
Thanks for the insight. I wasn't aware that `oid` could be missing as well.

reneleonhardt:
The richest corporation in the world had 14 years time to keep up with 14 new awesome Postgres versions since 2010... but they didn't want to? At least Amazon Aurora is current these days and saves a bit of the cost 😅
https://www.postgresql.org/about/news/postgresql-17-beta-1-released-2865/

Wouldn't ClickHouse be the modern choice if OLAP has to run on AWS? 🤔
https://clickhouse.com/blog/migrating-data-between-clickhouse-postgres
https://aws.amazon.com/solutions/implementations/clickhouse-cluster/

mp911de:

Issue #653, opened 2024-05-23
SSL SNI hostname with trailing dot unable to connect


seanmcnealy:
#### Versions

- Driver: n/a
- Database: n/a
- Java: n/a
- OS: n/a

#### Current Behavior

SSL SNI hostname with trailing dot unable to connect. Throws a bit of a weird error.

Disabling SNI works fine. So this is a minor issue. But the same hostnames worked in the previous version using default settings and do not work in the current version with default settings.

<details>
<summary>Stack trace</summary>

```
java.lang.RuntimeException: java.lang.IllegalArgumentException: Server name value of host_name cannot have the trailing dot
	at io.r2dbc.postgresql.client.ReactorNettyClient.registerSslHandler(ReactorNettyClient.java:422)
	at io.r2dbc.postgresql.client.ReactorNettyClient.lambda$connect$11(ReactorNettyClient.java:402)
	at reactor.netty.transport.TransportConfig$TransportChannelInitializer.initChannel(TransportConfig.java:418)
	at io.netty.channel.ChannelInitializer.initChannel(ChannelInitializer.java:129)
	at io.netty.channel.ChannelInitializer.handlerAdded(ChannelInitializer.java:112)
	at io.netty.channel.AbstractChannelHandlerContext.callHandlerAdded(AbstractChannelHandlerContext.java:1130)
	at io.netty.channel.DefaultChannelPipeline.callHandlerAdded0(DefaultChannelPipeline.java:609)
	at io.netty.channel.DefaultChannelPipeline.access$100(DefaultChannelPipeline.java:46)
	at io.netty.channel.DefaultChannelPipeline$PendingHandlerAddedTask.execute(DefaultChannelPipeline.java:1463)
	at io.netty.channel.DefaultChannelPipeline.callHandlerAddedForAllHandlers(DefaultChannelPipeline.java:1115)
	at io.netty.channel.DefaultChannelPipeline.invokeHandlerAddedIfNeeded(DefaultChannelPipeline.java:650)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.register0(AbstractChannel.java:514)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.access$200(AbstractChannel.java:429)
	at io.netty.channel.AbstractChannel$AbstractUnsafe$1.run(AbstractChannel.java:486)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask$$$capture(AbstractEventExecutor.java:173)
[...]

PR #656, opened 2024-06-06
#653 SNIHostName is going to throw an exception when hostname has a traili…

seanmcnealy:
…ng dot

Make sure that:

- [X] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.
- [X] You submit test cases (unit or integration tests) that back your changes.

 

#### Issue description

SSL SNI hostname with trailing dot unable to connect
 
#### Additional context

Minor issue, as there is an easy workaround to disable SNI through configuration that avoids the issue entirely. The underlying library throws an error when there is a trailing dot on an SNI hostname. Looks easy enough to match that library's validation in the SSLConfig validation.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/client/SSLConfig.java
+++ b/src/main/java/io/r2dbc/postgresql/client/SSLConfig.java
@@ -108,9 +108,7 @@ public final class SSLConfig {
                 return false;
             }
         }
-        if (input.endsWith("."))
-            return false;
-        return true;
+        return !input.endsWith(".");
     }
 
     //
--- a/src/main/java/io/r2dbc/postgresql/client/SSLConfig.java
+++ b/src/main/java/io/r2dbc/postgresql/client/SSLConfig.java
@@ -108,6 +108,8 @@ public final class SSLConfig {
                 return false;
             }
         }
+        if (input.endsWith("."))
+            return false;
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


mp911de:
How come that a host name returned from `InetSocketAddress` ends with a dot?

seanmcnealy:
Looks like some DNS clients support adding a trailing dot which signifies to not use a search domain when resolving a name. I'm not an expert on this, I just have a coworker who used this convention when setting some environment variables.

I've tested 1.0.4 (works) and 1.0.5 (throws validation exception) with the following code:

```kotlin

import io.r2dbc.postgresql.PostgresqlConnectionConfiguration
import io.r2dbc.postgresql.PostgresqlConnectionFactory
import io.r2dbc.postgresql.client.SSLMode

val config =
    PostgresqlConnectionConfiguration.builder()
        .host("database.internal.")
        .port(5432)
        .database("reporting")
        .username("user")
        .password("xxxx")
        .sslMode(SSLMode.REQUIRE)
        .build()
val factory = PostgresqlConnectionFactory(config)

factory.create().block()

```

I could also see trimming trailing dots before sending to the SNIHostName constructor as valid. That would support SNI better, I think. Unless changing hostnames at all can cause more surprising results.

mp911de:

Issue #332, opened 2020-09-19
How do I fetch COPY output result ?


unoexperto:
Is there official way of getting COPY output result ? I see that `CopyData` messages are coming in `PostgresqlResult.messages` but there is no way for me to get it because `PostgresqlResult.map` is hardcoded to get regular rows only.

Thank you.

mp911de:
There's no direct API for `CopyData` consumption. `PostgresReplicationStream` uses `CopyData` that lets you consume the replication stream of `START_REPLICATION SLOT … LOGICAL …`.

Care to elaborate on your use-case so we can understand what you want to achieve to discuss how to make it possible?

unoexperto:
@mp911de My use-case is data replication. So It's possible that single COPY query will return 100s of GB of data. As result I need to parse response in streaming fashion.

mp911de:
Data replication as in logical decode? How would you consume `COPY` though the pgjdbc driver? Maybe we can come up with something similar.

davecramer:
I read it as he just wants to consume the copy stream. Data replication is orthogonal to the request

unoexperto:
Dave is correct. I want to consume potentially unbounded stream of binary data. Typically COPY output is either CSV or BINARY. So my client code will take of parsing it.

What's also important is to know that the stream is fully consumed. For that I need to get `CopyDone`.

If I were to modify existing code with minimal intrusion I'd probably have something like `CopyDataStreaming` where `data` field is of `Flux<ByteBuf>` type.

mp911de:
Is that something we could map onto `PostgresqlResult` in the form of e.g.

```
/**
 * Return a stream of mapped objects from {@link CopyData}. The stream terminates when the underlying stream receives  {@link CopyDone}.
 */
<T> Flux<T> mapCopy(Function<CopyData, ? extends T> mappingFunction);
```

Since I'm not terribly familiar with `Copy` (and `CopyIn`, `CopyOut`), I'm happy for suggestions. Also an example how the calling code could look like would be helpful to find a suitable approach. `CopyData` holds a `ByteBuf` that can be consumed by the caller.

unoexperto:
Thank you, Mark. I'm afraid returning response in `ByteBuf` won't cut it. It can easily be terabyte of data. Let me fork and experiment with implementation options and I'll raise PR to discuss if solution is feasible.

mp911de:
Sure, happy to continue the discussion over the actual code. 

[...]

PR #720, opened 2026-07-01
Added COPY TO STDOUT support

jroper:
Provides #332

This adds a new `PostgresqlConnection.copyOut(String sql)` method to support `COPY ... TO STDOUT` statements.

In addition, PostgresqlResult now fails when executing these statements, rather than silently dropping all the data.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
+++ b/src/main/java/io/r2dbc/postgresql/PostgresqlConnection.java
@@ -16,6 +16,8 @@
 
 package io.r2dbc.postgresql;
 
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
 import io.r2dbc.postgresql.api.CopyInBuilder;
 import io.r2dbc.postgresql.api.ErrorDetails;
 import io.r2dbc.postgresql.api.Notification;
@@ -58,6 +60,7 @@ import java.time.Duration;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
+import static io.r2dbc.postgresql.PostgresqlStatement.WINDOW_UNTIL;
 import static io.r2dbc.postgresql.client.TransactionStatus.IDLE;
 import static io.r2dbc.postgresql.client.TransactionStatus.OPEN;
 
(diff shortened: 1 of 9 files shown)


mp911de:
Thanks a lot, like how this change is shaped. I am not sure when I will be able to have a look, will be on vacation starting next week (just fyi).

jroper:
One thing to note, I did consider originally adding this under `PostgresqlStatement.executeCopyOut`, because that would allow reusing the query parameter support from `PostgresqlStatement`, except then I learned that query parameters aren't supported in `COPY` statements by Postgres, the statement must be just plain SQL without parameters, so it made more sense to put it on `PostgresqlConnection`, especially given that that means it can mirror `copyIn`.

There's a few things that I'm not sure about - I explicitly complete the downstream sink when receiving a `CopyDone`. This is technically not needed since `CopyDone` is always followed by a `CommandComplete` which will always be followed by a stream complete by the upstream mechanics. I'm not sure what the consequences of one approach or the other are.

Something to be aware of, the Postgres protocol has no difficulties combining `COPY TO STDOUT` commands and other commands in a single query text. If we wanted to support this, we would need to add this support onto `PostgresResult`, eg, by adding a `mapCopyOut` method, so that a given result can be used as either. That's another way to implement this, one that allows more flexibility. It however doesn't mirror the copy in support, and I don't think there are any real use cases for mixing `COPY TO STDOUT` with any other commands, these are not transactional, it just doesn't make sense to me to do it.

I marked it as `@since 1.2`, under the assumption that this is a new API and therefore will warrant a minor version bump.

mp911de:
Sorry for the long silence. I appreciate where this change is going and its general design.

To make it more robust, we should handle an unintended `COPY IN` (e.g. `copyOut("COPY test FROM STDIN")`) by sending `CopyFail` when we receive `CopyInResponse`. Without responding to `CopyInResponse`, the connection gets stuck and becomes unusable. Looking at `PostgresqlResult`, the copy response handling is missing there as well, so this PR has already surfaced some valuable insights. Thanks for that.

I think two entrypoints for the typical copy in/out arrangements make sense. What is missing is a broader COPY API that is more generic, allowing to specify inputs as well as consuming Copy Out. That is clearly a second step. For now, I like an approach along the lines of the following design (captured here for documentation purposes, not to turn this ticket into a general COPY API discussion):

```java
interface CopyBuilder {

      CopyBuilder from(Publisher<ByteBuf> input);

      CopyBuilder fromMany(Publisher<? extends Publisher<ByteBuf>> input);

      Flux<PostgresqlCopyResult> build();
  }

  interface PostgresqlCopyResult {

      Mono<Long> getRowsUpdated();

      <T> Flux<T> map(BiFunction<ByteBuf, CopyMetadata, ? extends T> mapper);
  }
```

Usage:

```java
  // Import
  connection.copy(sql)
      .from(input)
      .build()
      .concatMap(PostgresqlCopyResult::getRowsUpdated);

  // Export
  connection.copy(sql)
      .build()
      .concatMap(result -> result.map(this::decode));
```

PR #724, opened 2026-07-02
Use non-blocking DNS resolution for client connections

### LINKED ISSUE #721: DefaultNameResolver blocks event loop during DNS resolution
We recently encountered 5-second connections stalls caused by DNS resolution. The problem is that such a stall blocks other ongoing requests on the same event loop. In Grafana, this situation looks like this:

Image

We attribute this to io.netty.resolver.DefaultNameResolver, which is blocking and is used by the driver (io.r2dbc.postgresql.client.BalancedResolverGroup). I wonder if it makes sense to replace it with io.netty.resolver.dns.DnsNameResolver, which is a non-blocking version of io.netty.resolver.InetNameResolver?

goutamadwant:
Make sure that:

#### Issue description

Fixes #721.

`ReactorNettyClient` used a custom resolver group that delegated hostname lookup to Netty's `DefaultNameResolver`, which can block the event loop during DNS resolution.

This switches `InetSocketAddress` connections to Reactor Netty's name resolver provider with round-robin selection enabled. That keeps the existing multi-address balancing behavior while using Reactor Netty's non-blocking DNS resolver.

#### New Public APIs

None.

#### Additional context

Validation:

- `./mvnw -Dtest=ReactorNettyClientUnitTests test`
- `./mvnw test`


src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
```diff
@@ -101,6 +102,10 @@ public final class ReactorNettyClient implements Client {
 
     private static final boolean DEBUG_ENABLED = logger.isDebugEnabled();
 
+    static void configureNameResolver(NameResolverProvider.NameResolverSpec nameResolverSpec) {
```
mp911de, line 105: Nit: Let's inline this as lambda

mp911de:

PR #727, opened 2026-07-08
Additionally support Postgis Geography codec under one generic PostgisCodec

### LINKED ISSUE #528: Add geography data type support
## Feature Request
Add geography data type support

"geography" data type can already be encoded to / decoded from the same Geometry abstract class as "geometry".
I tried copying the geometry codec and renaming it, and it worked.

The new codec can be included inside BuiltinDynamicCodecs as the geometry one.

mdedetrich:
Make sure that:

- [X] You use the code formatters provided [here](https://github.com/pgjdbc/r2dbc-postgresql/blob/master/intellij-style.xml) and have them applied to your changes. Don't submit any formatting related changes.
- [X] You submit test cases (unit or integration tests) that back your changes.

 

#### Issue description

Resolves: https://github.com/pgjdbc/r2dbc-postgresql/issues/528

 
#### New Public APIs

`PostgisGeographyCodec` (if this counts as an API).

#### Additional context

Note that the implementation does have a fair bit of boilerplate due to `PostgisGeographyCodec` being final and hence not being able to be subclassed. One alternative would be to have a package private implementation that both `PostgisGeographyCodec` and `PostgisGeometryCodec` can share (that way both `PostgisGeographyCodec` and `PostgisGeometryCodec` can be final).

As suggested in the linked github issue, the implementation is just a copy paste of `PostgisGeometryCodec` that also uses the `org.locationtech.jts.geom.Geometry` but this allows you to decode postgres `geography` types.

### DIFF
--- a/src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
+++ b/src/main/java/io/r2dbc/postgresql/codec/BuiltinDynamicCodecs.java
@@ -34,8 +34,6 @@ import java.util.stream.Collectors;
  */
 public class BuiltinDynamicCodecs implements CodecRegistrar {
 
-    private static final Object EMPTY = new Object();
-
     enum BuiltinCodec {
 
         HSTORE("hstore"),
@@ -47,7 +45,18 @@ public class BuiltinDynamicCodecs implements CodecRegistrar {
             public boolean isSupported() {
                 return this.jtsPresent;
             }
-        }, VECTOR("vector");
+        },
+        POSTGIS_GEOGRAPHY("geography") {
+
+            private final boolean jtsPresent = isPresent(BuiltinDynamicCodecs.class.getClassLoader(), "org.locationtech.jts.geom.Geometry");
+
+            @Override
+            public boolean isSupported() {
+                return this.jtsPresent;
+            }
+        },
+
+        VECTOR("vector");
 
         private final String name;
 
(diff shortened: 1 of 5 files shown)


mdedetrich:
After realizing an issue with this implementation, specifically that encoding an object using `Codec.canEncode(Object value)`, would have gotten surprising results into how that object would have been encoded depending on what order `DefaultCodecLookup.findEncodeCodec` finds them due to having multiple codecs sharing the JTS `Geography` type.

Given that, I opted for a design that is more semantically correct, i.e. treat `Geography` the same way that `String` is treated and make a generic `PostgisCodec` that supports both geography and geometry in the same way `String` supports multiple types (`VARCHAR`, `TEXT`, `NAME` etc etc).

This did require touching some other internals i.e. `BuiltinDynamicCodecs` had to be updated to support retrieving multiple oid's from multiple rows and join them into a single codec (currently only supported one oid per row), but I think the end result is cleaner, better and more idiomatic as just like `String`, the only difference between `Geography` and `Geometry` are internal differences in how Postgres calculates distances, but they both fundamentally represent the same type on a client.

mdedetrich:
@mp911de When do you have capacity to review this PR? I am asking because there are follow up pull requests based on this one that I want to look at.

mp911de:
This is an interesting addition, but the changes touch several places, apparently to avoid registering two codec instances. Honoring the method-provided OID for encoding is a useful improvement on its own right. However, discovery and type translation now have two owners, making the built-in registration mechanism harder to follow.

`DiscoveredType`, `collectList()`, and `registerCodecs()` introduce an intermediate discovery step merely to pair two OIDs. I suggest registering the codec twice, once for each OID, while preserving geometry as the default encoding type to avoid further ceremony. As a separate optimization, `PostgisGeometryCodec.geometryFactory` could be made static.

A few details could also be simplified: The explanatory comments are rather lengthy for the behavior involved with causing noise only. Such changes lead to overall code quality deterioration as comments vs. behavior diverges over time. Tests calling `registerCodecs()` directly bypass discovery and tie coverage to the intermediate representation introduced here.

We appreciate community contributions, but we do need them to meet a reasonable baseline for consideration. If LLM-generated code still requires substantial simplification, submitting it largely shifts that work to the maintainers resulting in negative value contributions. There's little difference in maintainers prompting an LLM and then cleaning up AI-generated code bloat vs. someone else prompting an LLM and then maintainers cleaning up AI-generated code bloat.

Issue #242, opened 2020-02-04
High CPU usage when cancelled query


steven-sheehy:
## Bug Report

Using 0.8.1 in our [project](https://github.com/hashgraph/hedera-mirror-node/blob/master/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/service/TopicMessageServiceImpl.java) and running into an issue where we see very high CPU utilization after performing a simple select on a table. The below jstack output shows some sort of infinite loop in `io.r2dbc.postgresql.client.ReactorNettyClient$BackendMessageSubscriber.drainLoop(ReactorNettyClient.java:828)`.

Also, after the query is completed and returned the appropriate results, `pg_stat_activity` still shows that query as active even though it should've been completed and returned to the pool. I've confirmed with logs for doOnCancel and doOnComplete that this flux is actually complete. Eventually all connections in the thread pool show as active and the thread pool is exhausted. It's not clear if these are related or two separate issues.

#### Versions

- Driver: 0.8.1
- Database: PostgreSQL 9.6 (Google Cloud SQL)
- Java: OpenJDK 11
- OS: Ubuntu 18.04

#### Current Behavior

[Application logs](https://github.com/r2dbc/r2dbc-postgresql/files/4151407/grpc.log)

<details>
<summary>jstack output</summary>

```
2020-02-03 22:11:34
Full thread dump OpenJDK 64-Bit Server VM (11.0.6+10-post-Ubuntu-1ubuntu118.04.1 mixed mode, sharing):

Threads class SMR info:
_java_thread_list=0x00007fd974001880, length=20, elements={
0x00007fd9a0165000, 0x00007fd9a0167000, 0x00007fd9a016f000, 0x00007fd9a0171800,
0x00007fd9a0173800, 0x00007fd9a0175800, 0x00007fd9a01ac800, 0x00007fd9a01b1800,
0x00007fd9a0bb1800, 0x00007fd9a046a000, 0x00007fd9a0997000, 0x00007fd9a0997800,
0x00007fd9a0c36000, 0x00007fd9a15d9800, 0x00007fd9a15f0000, 0x00007fd9a1872800,
0x00007fd9a187d800, 0x00007fd9a0015000, 0x00007fd9640a5000, 0x00007fd95400b800
}

[...]

PR #731, opened 2026-07-25
Release results that are discarded after cancellation

### LINKED ISSUE #661: `ReactorNettyClient` stuck on cancelled Conversation
#### Versions

- Driver: 1.0.5
- Database: PostgreSQL 13.12
- Java: 17
- OS: MacOS, Linux

#### Current Behavior

When you have query zipped in parallel with some other failed function and that query return more than 256 rows it can leads to the case when you no have real consumer, because chain was cancelled, but you receive data from database that start to save it to ReactorNettyClinet.buffer.
When this happens, any other attempts to get data from the database will fail because ReactorNettyClient.BackendMessageSubscriber.tryDrainLoop never call drainLoop because stucked conversation no have demands
```
private void tryDrainLoop() {
    while (hasBufferedItems() && hasDownstreamDemand()) {
        if (!drainLoop()) {
            return;
        }
    }
 }
```

Can reproduce using https://github.com/agorbachenko/r2dbc-connection-leak-demo
If you increase System property "reactor.bufferSize.small" to 350, the attached example will start working

agorbachenko:
Make sure that:

- [x] Issue referenced: #661 (and r2dbc/r2dbc-pool#198, where this was originally reported).
- [x] No formatting-related changes; the diff follows the surrounding style.
- [x] Integration test included, which fails before and passes after the change.

#### Issue description

Resolves #661, and the connection leak reported in r2dbc/r2dbc-pool#198.

`PostgresqlStatement.execute()` decorates the `Flux<PostgresqlResult>` with `Operators.discardOnCancel`. After a cancellation that operator keeps the subscription and drops any further result through `Operators.onDiscard`. `PostgresqlResult` is reference-counted and drains its remaining messages in `deallocate()`, so a discarded result has to be released to be drained.

The `doOnDiscard(ReferenceCounted.class, ReferenceCountUtil::release)` that is already present is registered *above* the `map(…)`, so it is not part of `discardOnCancel`'s subscriber context and never applies to the mapped results. A result that is cancelled before it is consumed is therefore dropped without being released, and its `windowUntil` window is never drained.

From that point the connection is unusable:

* the `Conversation` stays at the head of `BackendMessageSubscriber.conversations` with `demand == 0`,
* `drainLoop()` leaves at the head because the head has no demand (intentional, #242),
* `demandMore()` cannot request from the transport while `hasBufferedItems()` is `true`,
* every conversation queued behind it starves.

State of the message subscriber once wedged, read from a running reproducer:

```
conversations=2 buffer=59 upstreamDemand=171 terminated=false
head conversation: demand=0 sinkCancelled=false
first buffered message=DataRow
```

Note `sinkCancelled=false`: the client never observes a cancellation, it only sees a consumer that stopped requesting. That is why the investigations in r2dbc/r2dbc-pool#198 could not pin this on reactor-pool or on `usingWhen` — at or below the client this is indistinguishable from a slow consumer, so no change keyed on cancellation inside `ReactorNettyClient` could have fixed it.

With a pool the symptom is a leaked connection: the cancelling borrower releases it, and the next borrower's query never completes even though `pg_stat_activity` shows the backend `idle` with the response fully sent.

The handler is **moved** rather than duplicated. `Operators.discardLocalAdapter` composes handlers (`safeConsumer.andThen(consumer)`), so registering a second one would give every operator upstream of both registrations both handlers, and the rows `FluxWindowPredicate` discards would be released twice. Measured over the 940 pre-existing tests in `PostgresCancelIntegrationTests`: 0 `Error in discard hook` / `IllegalReferenceCountException` unpatched, 80 with a second handler added, 0 with the handler moved. Since the discard context propagates upstream only, one registration below `discardOnCancel` covers both concerns — `WindowPredicate` still sees it, and now `discardOnCancel` does too.

#### New Public APIs

None.

#### Additional context

The new test, `PostgresCancelIntegrationTests#shouldReuseConnectionWhenResultIsDiscardedAfterCancel`, cancels a streaming response without consuming the result and then reuses the connection. It is deterministic — no pool and no race involved: 5/5 repetitions fail before the change (the second query times out), 5/5 pass after.

Also verified:

* full unit suite: 1408 tests, 0 failures,
* `-Dtest='*IntegrationTests'`: 1267 tests, 0 failures, 11 skipped, including the 946 existing cancellation repetitions,
* the reproducer from r2dbc/r2dbc-pool#198 (Spring Data R2DBC, pool of 1, query `zipWith` a failing `Mono`), upgraded to Spring Boot 4.0.7 / reactor-core 3.8.6 / reactor-pool 1.2.6 / Java 25: wedges after about three queries on both 1.1.2.RELEASE and `main`, and survives 127 queries with no stall and no leak reports with this change,
* a racing bad-actor/good-actor harness against a real connection, based on @chemicL's r2dbc-lab with the mocks replaced by the actual driver: stalls within 15 iterations on both 1.1.2.RELEASE and `main`, and passes 150 iterations × 2 response sizes × 2 scenarios with this change.

Notes:

* This is not the nested-`flatMap` deadlock discussed at the end of #661. That one is a consumer that never requests; this one is a cancellation that loses the obligation to drain.
* If you prefer a different placement, `FluxDiscardOnCancelSubscriber.onNext` could release `ReferenceCounted` items it drops instead of relying on a context handler. I went with the smaller, local change.
* The other `windowUntil` usage, `DefaultPostgresqlReplicationConnection.startReplication()`, is not affected. That chain has no `discardOnCancel`, so a cancellation propagates to the exchange normally and the conversation terminates; and `PostgresReplicationStream` is not reference-counted, so there is nothing a discard handler could release. `LogicalDecodeIntegrationTests` passes with `-Dio.netty.leakDetection.level=paranoid` and no discard-hook errors.


chemicL (approved):
LGTM!

agorbachenko:
Thanks for the review, @chemicL! @mp911de, this is ready whenever you have a moment.

mp911de:
Thanks a lot @agorbachenko. I am a bit overwhelmed by the change but given that the smallest changes are typically the ones with the most impact, it shouldn't be a surprise in hindsight. 

You were able to resolve the nasty Reactor Pool issue that only showed when using Postgres. I am always impressed by how much AI tools are able to help with working their way to diagnose the underlying cause. @chemicL and I spent several days trying to figure out what causes the issue surfacing through the Connection Pool but ultimately, weren't able to get to the root cause. 

Looking at the other places we use `discardOnCancel`, do you mind cross-checking whether these are also subject to similar problems?

agorbachenko:
> Looking at the other places we use `discardOnCancel`, do you mind cross-checking whether these are also subject to similar problems?

Actually, I've already did this research. It showed that other usages were unaffected since they didn't meet the conditions. Unfortunately, the details of that research have been lost, and I can't provide them as evidence.

mp911de:
Thanks a lot. It has been a thought that crossed my mind. With you already having finished the research, that’s all I wanted to know for a bit more peace of mind.
Am 17. Sept. 2026, 12:33 +0200 schrieb Alexei Gorbachenko ***@***.***>:

PR #736, opened 2026-08-29
Release bind parameters of unwritten messages.

### LINKED ISSUE #734: Bind parameter ByteBufs leak when a statement is cancelled before it is flushed
### Versions

| | |
|---|---|
| `org.postgresql:r2dbc-postgresql` | `1.1.2.RELEASE` (current latest) |
| `io.projectreactor:reactor-core` | `3.7.19` |
| `io.netty:netty-buffer` | `4.1.136.Final` |
| JVM | Temurin 25.0.4+7 (LTS) |
| PostgreSQL | 16 (via Testcontainers) |

### Summary

When a reactive chain containing a `PostgresqlStatement` is **cancelled before that statement is
flushed to the server**, the `CompositeByteBuf` holding its bind parameters is never released.
Netty's leak detector attributes it to `PostgresqlStatement.collectBindingParameters`.

This is the **request** side. It is distinct from the previously reported response-side leaks
(`DataRow` / `BackendMessageDecoder`, e.g. #580), and it reproduces only under cancellation rather
than under load.

### Stack trace

Captured with `-Dio.netty.leakDetection.level=paranoid -Dio.netty.leakDetection.targetRecords=32`.
Frames below `MonoReduceSeed` are application code and are elided.

```text
LEAK: ByteBuf.release() was not called before it's garbage-collected.
Recent access records:
#1:
	io.netty.buffer.AdvancedLeakAwareByteBuf.order(AdvancedLeakAwareByteBuf.java:71)
	io.netty.buffer.CompositeByteBuf.newComponent(CompositeByteBuf.java:346)
	io.netty.buffer.CompositeByteBuf.addComponent0(CompositeByteBuf.java:287)
	io.netty.buffer.CompositeByteBuf.addComponent(CompositeByteBuf.java:265)
	io.netty.buffer.CompositeByteBuf.addComponent(CompositeByteBuf.java:222)
	io.r2dbc.postgresql.PostgresqlStatement.lambda$collectBindingParameters$1(PostgresqlStatement.java:286)
	reactor.core.publisher.MonoReduceSeed$ReduceSeedSubscriber.onNext(MonoReduceSeed.java:116)
	...
```

### How to reproduce

Any chain that issues several parameterised statements and is cancelled partway through. Shape:

```java
// CREATE TABLE t (id uuid PRIMARY KEY, n int NOT NULL, s varchar(40) NOT NULL);

Mono<Void> work = Flux.range(0, 40)
    .concatMap(i -> Mono.from(connection
            .createStatement("INSERT INTO t (id, n, s) VALUES ($1, $2, $3)")
            .bind("$1", UUID.randomUUID())
            .bind("$2", i)
            .bind("$3", "row-" + i)
            .execute())
        .flatMap(result -> Mono.from(result.getRowsUpdated())))
    .then();

StepVerifier.create(work)
    .thenAwait(Duration.ofMillis(N))   // sweep N across the duration of the chain
    .thenCancel()
    .verify();
```

Run the suite with the paranoid leak detector enabled. Sweeping the cancellation point across the
lifetime of the chain (we used 16 points) makes it land reliably; a single fixed delay is flaky
because the window is narrow.

Two properties worth noting for anyone trying to confirm this:

1. **The detector samples, and reporting requires a GC after the buffer is orphaned.** The same tree
   produced one red run and two green ones for us. It is not deterministic, so absence of a report in
   a single run is not evidence of absence.
2. **The assertion is the leak gate, not the test.** All test cases pass; only the leak detector
   fires.

### Why a consumer-side workaround does not work

We tried to release it from outside the driver with a `ConnectionFactory` decorator that installs a
Reactor `doOnDiscard(ByteBuf.class, ...)` handler on the subscription. It does not work, and the
reason looks relevant to the fix:

- The decorator was confirmed installed (its `create()` ran) and the leak still reproduced in that
  same run.
- The discard handler was invoked **zero** times.

Tracing it: the buffer produced by `ByteBufUtils.encode` is *delivered* downstream into the
`FluxConcatMapNoPrefetch` that assembles the `Bind` message, and the assembled message is then
dropped unwritten when the chain is cancelled. A value that was successfully delivered downstream is
never routed through Reactor's discard hook, so `doOnDiscard` cannot see it. We measured the three
relevant shapes to confirm this is Reactor's contract and not a mistake on our side:

| shape | value routed to `onDiscard`? |
|---|---|
| rejected by `filter` | yes |
| `delayElement` cancelled before emit | yes |
| `publishOn` queue cleared on cancel | **no** |

So the release has to happen inside the driver, at whatever point owns the assembled `Bind` message
when the subscription goes away.

### Impact

Cancellation is a normal condition for a server that accepts HTTP requests: a client disconnect or a
proxy timeout tears the chain down. In a long-lived process, each cancelled chain leaks the bind
buffers of every statement that had been encoded but not yet flushed — in the batch above, up to 40
statements per cancellation. That is off-heap memory that is never reclaimed.

There is a secondary cost for anyone running the leak detector in CI: because the detector samples,
an affected build turns red intermittently and the failure surfaces in whichever test happened to be
running when the GC occurred, not in the one that caused it.

### Related

- #393 — same area (parameter encoding), but a single unexplained occurrence on 0.8.7 with no
  identified mechanism. This report adds the mechanism (cancellation before flush), a reproduction
  shape, and current versions.
- #580 — also a `ByteBuf` leak, but on the response side under load.

### Caveat

The attribution is the **allocation site**, not a confirmed ownership root cause: the trace says
where the buffer was created, not who was supposed to release it.

SecondPort:
Make sure that:

#### Issue description

Resolves #734.

`ExtendedFlowOperator.getMessages(…)` retains the bound parameter buffers for the `Bind` it creates, and that reference is given back only by the `finally` block in `Bind.encode(ByteBuf)`. `ExtendedFlowOperator.close(…)` returns the operator's own reference from `doFinally`. A buffer therefore reaches `refCnt` 0 only if the message is also encoded, and any message that never reaches the wire strands one reference per non-null parameter.

Two paths do that:

1. **The conversation is refused before the request stream is subscribed.** The three `fetch*` methods pass `Flux.just(new CompositeFrontendMessage(factory.createMessages()))` as an argument, so the messages — and their retains — are created while the exchange is assembled. When `Client.exchange(…)` refuses on `!isConnected()`, the request stream is never subscribed, `doFinally(operator.close)` returns one reference and the `Bind`'s reference is stranded.

2. **The channel dies while the message is queued.** `ChannelOperations.send(Publisher, Predicate)` returns `Mono.error(AbortedException.beforeSend())` on its `!channel().isActive()` branch without referencing `dataStream`. Since `message.encode(alloc)` is a cold `Mono.fromSupplier`, it is never subscribed, so `Bind.encode(ByteBuf)` and its `finally` never run. The sibling `sendObject(Object)` calls `ReactorNetty.safeRelease(message)` under the same check — reactor-netty releases what it holds by value, but it cannot release a cold publisher it never subscribed, so ownership stays with the caller.

This is the same class of request-side leak as the `CopyData` fix in 68694fc7, one level up: `FrontendMessage.dispose()` already documents that a message discarded before reaching the wire must release its buffers and that disposal is idempotent. These two paths simply did not honour it.

**A correction to the issue report, which I filed:** a bare cancel over a *live* connection does **not** leak. `Operators.discardOnCancel` does not cancel upstream — it keeps draining until the conversation terminates, so `doFinally(operator.close)` fires late and `Bind.encode`'s `finally` has already run by then. Measured: `refCnt` is still 1 immediately after `thenCancel()` and 0 only after the connection closes. The leak requires the message never to be **encoded**; cancellation is the common real-world trigger, because a cancelled chain releases or closes the connection. The issue title is imprecise on that point.

Also worth noting: `Client.send(FrontendMessage)` does **not** leak and is unchanged. `doSendRequest` wraps the message in `Mono.just(…)`, so `concatMap` treats it as a scalar and `WeakScalarSubscription.cancel()` re-discards it into the existing `doOnDiscard(FrontendMessage.class, FrontendMessage::dispose)`. The `exchange(…)` shape is a non-scalar `FluxConcatArray`, where an already-emitted value is not re-discarded.

#### New Public APIs

None. No signature changes, no new configuration.

#### Additional context

The change is +12/-4 across two files:

* `ExtendedFlowDelegate` — create the messages upon subscription (`Mono.fromSupplier`) instead of while assembling the exchange, at the three `fetch*` sites. Nothing is retained for a request stream that is never subscribed, which covers the `exchange(…)` refusal, `addConversation`'s subscribe-time re-check and a full conversation queue in one place.
* `ReactorNettyClient` — dispose the message once the write attempt terminated. `dispose()` is idempotent and a no-op for messages that got encoded (`Bind` guards on its `AtomicBoolean`, `CopyData` on `refCnt() > 0`, `CompositeFrontendMessage` propagates, the default is a no-op).

I deliberately did not collapse the double ownership by dropping the extra `retain()` in `getMessages(…)`: `operator.close(…)` fires from `doFinally` and can run while the `Bind` is still queued, which would hand released buffers to `Bind.encode(ByteBuf)`.

Tests are **unit** tests rather than integration tests, because PR CI runs with `-D skipITs` and a regression test under `*IntegrationTests` would not guard this. Both new classes drive a real `ReactorNettyClient` over a loopback socket — no Docker, no PostgreSQL — and assert `refCnt()` directly rather than relying on the sampling leak detector.

* `ReactorNettyClientUnitTests` (the class removed in 86ab56d5, reinstated) — path 2, plus a control test on a live channel.
* `PostgresqlStatementCancellationUnitTests` — both paths end to end through `PostgresqlStatement`, plus a control test that cancels over a live connection.

The control tests pass without the fix; they are what shows the failing assertions are the bug rather than a harness artefact. Verified on this branch: without the src/main changes the three regression tests fail with `expected: 0 but was: 1` and the two control tests pass; with them, `./mvnw -B verify` reports 1413 unit tests green.


mp911de:
We appreciate community contributions, but we do need them to meet a reasonable baseline for consideration.

> [!TIP]
> This means:
> * Expressing your intent with a concise and reasonable description

When issues require significant effort from our side to initially understand your request, we have to deprioritize or even reject them in favor of tickets that are ready for discussion as we have to be mindful of our limited time, and we need to focus it where it has the most impact. If you would like us to spend some time working on this issue, please spend some time making this a substantial and quality ticket description.

When describing your concern, please focus on the aspects that matter. There's no need to duplicate parts of our code in the description or use excessive summarization or bullet lists. Instead, use rather links if you want to point out subtle details that matter in the specific context. Avoid conflicting requests such as asking for guidance and then answering questions yourself, as such descriptions neglect raising a ticket in the first place.

src/main/java/io/r2dbc/postgresql/client/ReactorNettyClient.java
```diff
@@ -200,7 +200,9 @@ private ReactorNettyClient(Connection connection, ConnectionSettings settings) {
                 if (DEBUG_ENABLED) {
                     logger.debug(this.context.getMessage(String.format("Request:  %s", message)));
                 }
-                return connection.outbound().send(message.encode(this.byteBufAllocator));
+                // send(...) does not subscribe the encode publisher if the channel is no longer active, hence dispose the message once the write attempt terminated.
+                // dispose() is idempotent and a no-op for messages that got encoded.
+                return connection.outbound().send(message.encode(this.byteBufAllocator)).then().doFinally(ignore -> message.dispose());
```
mp911de, line 205: So if `send(…)` does not subscribe to the publisher, how is `doFinally(…)` supposed to work? Delaying buffer creation (as in `Mono.fromSupplier(…)`) or in the typical `FrontendMessage` implementations makes sense. Care to elaborate how `doFinally(…)` is then being invoked if `send(…)` no longer subscribes?

PR #738, opened 2026-09-04
Avoid per-parameter CompositeByteBuf when collecting bind parameters

### LINKED ISSUE #735: Avoid per-parameter `CompositeByteBuf` and Reactor pipeline in `collectBindingParameters`
`collectBindingParameters` reduces every non-null parameter into an `Unpooled.compositeBuffer()`, although almost every codec emits exactly one `ByteBuf` and `Bind.encode` flat-copies the bytes with `writeBytes`. 

Each parameter pays for an unpooled composite, a `reduce` subscriber and a `concatMap` inner subscription that add nothing to the encoded message.

Proposal: pass single-buffer parameters through as-is and only build a composite when a publisher emits more than one buffer.

inemtsev:
Make sure that:

#### Issue description

Resolves #735.

`collectBindingParameters` reduced every parameter into an unpooled `CompositeByteBuf`, although almost every codec emits a single `ByteBuf` that `Bind.encode` copies anyway. `Mono` sources now pass their buffer through as-is. Other sources reduce into a composite from the channel allocator with no component limit. Wire format, NULL handling and release semantics are unchanged.

#### New Public APIs

* `ByteBufUtils.combine(Publisher<? extends ByteBuf>, ByteBufAllocator)`

#### Additional context

JMH, one 4-byte buffer per parameter, `-prof gc`, JDK 21, Apple M4 Pro:

| params | ns/op before / after | B/op before / after |
|-------:|---------------------:|--------------------:|
| 1      | 170 / 140            | 994 / 810           |
| 5      | 742 / 643            | 2937 / 2016         |
| 10     | 1284 / 1039          | 5361 / 3521         |


inemtsev:
@mp911de Let me know what you think.

src/main/java/io/r2dbc/postgresql/util/ByteBufUtils.java
```diff
@@ -59,4 +64,28 @@ public static ByteBuf encode(ByteBufAllocator byteBufAllocator, CharSequence s)
         return byteBuf;
     }
 
+    /**
+     * Combine the buffers emitted by the given {@link Publisher} into a single {@link ByteBuf}, taking ownership of the emitted buffers. A {@link Mono} source emits at most one
+     * buffer and is passed through without aggregation machinery (issue #735); other sources are aggregated into a {@link CompositeByteBuf}. An empty source yields a zero-length
```
mp911de: Pardon me but since when do we mention ticket numbers in production code comments?

src/main/java/io/r2dbc/postgresql/util/ByteBufUtils.java
```diff
@@ -59,4 +64,28 @@ public static ByteBuf encode(ByteBufAllocator byteBufAllocator, CharSequence s)
...
+        Assert.requireNonNull(buffers, "buffers must not be null");
+        Assert.requireNonNull(allocator, "allocator must not be null");
+
+        if (buffers instanceof Mono) {
+            return ((Mono<ByteBuf>) buffers)
+                .defaultIfEmpty(Unpooled.EMPTY_BUFFER)
+                .map(buffer -> buffer != Unpooled.EMPTY_BUFFER ? buffer : allocator.buffer(0));
+        }
+
+        return Flux.from(buffers).<ByteBuf>reduceWith(allocator::compositeBuffer, (composite, buffer) -> ((CompositeByteBuf) composite).addComponent(true, buffer));
```
mp911de: That's an interesting change but we need to be careful regarding quadratic copying for larger buffers (`compositeBuffer()` with Netty's default of `16` components): Netty consolidates all components (full copy) every time the count exceeds `16`.

src/main/java/io/r2dbc/postgresql/util/ByteBufUtils.java
```diff
@@ -59,4 +64,28 @@ public static ByteBuf encode(ByteBufAllocator byteBufAllocator, CharSequence s)
...
+     */
+    @SuppressWarnings("unchecked")
+    public static Mono<ByteBuf> combine(Publisher<? extends ByteBuf> buffers, ByteBufAllocator allocator) {
+        Assert.requireNonNull(buffers, "buffers must not be null");
+        Assert.requireNonNull(allocator, "allocator must not be null");
+
+        if (buffers instanceof Mono) {
+            return ((Mono<ByteBuf>) buffers)
+                .defaultIfEmpty(Unpooled.EMPTY_BUFFER)
+                .map(buffer -> buffer != Unpooled.EMPTY_BUFFER ? buffer : allocator.buffer(0));
```
mp911de, line 87: What's the reason to request `Unpooled.EMPTY_BUFFER` when the buffer is already `Unpooled.EMPTY_BUFFER` (`buffer(0)` returns `EmptyByteBuf`)?

mp911de (changes requested):
Thank you for your contribution. However, I notice several aspects that suggest this change hasn't been thoroughly engineered before submission. The changes appear to lack context for our specific codebase patterns, and there are inconsistencies that would have been caught otherwise.
We appreciate community contributions, but we do need them to meet a reasonable baseline for consideration.

When contributions require significant rework from our side, we have to deprioritize or even reject them in favor of changes that are ready for integration as we have to be mindful of our limited time, and we need to focus it where it has the most impact. If you would like us to spend some time working on this pull request, please spend some time making this a substantial and quality contribution.

src/main/java/io/r2dbc/postgresql/util/ByteBufUtils.java
inemtsev: Good idea, the old code `Unpooled.compositeBuffer()` had this limit also, but Integer.MAX_VALUE is more defensive so I will apply it here.

src/main/java/io/r2dbc/postgresql/util/ByteBufUtils.java
inemtsev: Just checked, buffer(0) resolves to `heapBuffer(0, Integer.MAX_VALUE)`. The allocator only returns `EmptyByteBuf` when both capacities are zero. The swap is needed because Bind.encode treats Unpooled.EMPTY_BUFFER by identity as SQL NULL, so an empty parameter must be a distinct buffer or it is sent as NULL. I'll add a comment in code.

src/main/java/io/r2dbc/postgresql/util/ByteBufUtils.java
inemtsev: Removed

mp911de:

